PR #2, opened 2024-09-10
Key2ds

tertsdiepraam:


src/commands/key2ds.rs
```diff
@@ -0,0 +1,166 @@
...
+
+    /// use SHA1 for the DS hash
+    #[arg(short = '1', overrides_with_all = ["one", "two", "four", "algorithm"])]
+    one: bool,
+
+    /// use SHA256 for the DS hash
+    #[arg(short = '2', overrides_with_all = ["one", "two", "four", "algorithm"])]
+    two: bool,
+
+    /// use SHA384 for the DS hash
```
mozzieongit, line 36: Maybe change the algorithm names to be the mnemonics, so that the algorithms in the help output and the arguments for `--algorithm` are the same: `SHA-1`, `SHA-256`, and `SHA-384`. I got confused for a moment when trying to test the algorithms ^^

src/commands/key2ds.rs
```diff
@@ -0,0 +1,166 @@
...
+        // Otherwise we try to determine a similar hash to the key
+        match sec_alg {
+            SecAlg::RSASHA256
+            | SecAlg::RSASHA512
+            | SecAlg::ED25519
+            | SecAlg::ED448
+            | SecAlg::ECDSAP256SHA256 => DigestAlg::SHA256,
+            SecAlg::ECDSAP384SHA384 => DigestAlg::SHA384,
+            SecAlg::ECC_GOST => DigestAlg::GOST,
+            _ => DigestAlg::SHA1,
```
mozzieongit, line 163: Is using SHA-1 as the fallback a good idea? If new algorithms were to get introduced and we forget to update the above list, they would fallback to using SHA-1, right?

mozzieongit (changes requested):

src/commands/mod.rs
```diff
@@ -11,6 +12,13 @@ pub enum Command {
     #[command(name = "nsec3-hash")]
     Nsec3Hash(self::nsec3hash::Nsec3Hash),
 
+    /// Generate a DS RR from the DNSKEYS in keyfile
+    ///
+    /// The following file will be created: `K<name>+<alg>+<id>.ds`
+    /// The base name `(K<name>+<alg>+<id>` will be printed to stdout
```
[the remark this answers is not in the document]
tertsdiepraam: This was copied from `ldns-key2ds` so I assumed it had some meaning before I implemented it. But yeah, probably should be removed.

src/commands/key2ds.rs
tertsdiepraam: These arguments are only there for compatibility. So we're out of luck here :)

We could (later) refactor this with separate argument parsing (like I did for notify).

src/commands/key2ds.rs
tertsdiepraam: It's the current ldns behaviour: https://github.com/NLnetLabs/ldns/blob/b87b8e1d212b43f225891570ebc5045c6331dd8a/examples/ldns-key2ds.c#L48C1-L48C14

If we want to change it, let's take it to a separate PR, where we can discuss this in depth.

src/commands/key2ds.rs
mozzieongit: Yes, the arguments need to stay for compatibility. I was only confused by the help output saying the algorithm is spelled `SHA1`, and then tried using it with `--algorithm` while testing. The mnemonics for the hash algorithms for DS digests are with dashes (e.g. SHA-1) (while the DNSKEY algorithms don't use delimiters (e.g. RSASHA1)).
I'm just trying to avoid confusion for people trying to migrate to using `--algorithm` or `-a`.

```suggestion
    /// use SHA-1 for the DS hash
    #[arg(short = '1', overrides_with_all = ["one", "two", "four", "algorithm"])]
    one: bool,

    /// use SHA-256 for the DS hash
    #[arg(short = '2', overrides_with_all = ["one", "two", "four", "algorithm"])]
    two: bool,

    /// use SHA-384 for the DS hash
```

src/commands/key2ds.rs
mozzieongit: Ok

src/commands/mod.rs
```diff
@@ -11,6 +12,14 @@ pub enum Command {
     #[command(name = "nsec3-hash")]
     Nsec3Hash(self::nsec3hash::Nsec3Hash),
 
+    /// Generate a DS RR from the DNSKEYS in keyfile
+    ///
+    /// The following file will be created for each key:
+    /// `K<name>+<alg>+<id>.ds`.The base name `K<name>+<alg>+<id>`
```
mozzieongit, line 18: Got another typo :P
```suggestion
    /// `K<name>+<alg>+<id>.ds`. The base name `K<name>+<alg>+<id>`
```

src/commands/key2ds.rs
tertsdiepraam: Ah gotcha! Will fix

mozzieongit (approved):

src/commands/key2ds.rs
```diff
@@ -0,0 +1,166 @@
...
+            format!(
+                "Failed to open public key file \"{}\": {e}",
+                self.keyfile.display()
+            )
+        })?;
+        let zonefile = domain::zonefile::inplace::Zonefile::load(&mut file).unwrap();
+        for entry in zonefile {
+            let entry = entry.map_err(|e| {
+                format!(
+                    "Error while reading public key file \"{}\": {e}",
```
ximon18, line 79: This seems a strange error to me. There was a problem iterating over a zone file but the error is about reading a public key file?

src/commands/key2ds.rs
```diff
@@ -0,0 +1,166 @@
...
+            let Entry::Record(record) = entry else {
+                continue;
+            };
+
+            let class = record.class();
+            let ttl = record.ttl();
+            let owner = record.owner();
+
+            // Of the records that we see, we only care about DNSKEY records
+            let ScannedRecordData::Dnskey(dnskey) = record.data() else {
```
ximon18, line 94: You could also just test against record.rtype() and above the class, ttl, owner block too.

src/commands/key2ds.rs
mozzieongit: key2ds is usually used on key files containing only a single dnskey (e.g. generated by keygen). These files are in zonefile format, and you could use a single zone file containing all dnskeys you want to get DS records from. So maybe saying "Error while reading public key *from* file" will keep it ambiguous yet clear enough to apply to both reading from an actual zone file (used as a normal zone file), and reading from a keyfile (that is in zone file format)?

ximon18: Ah!

tertsdiepraam: I like @mozzieongit's suggestion. Will push a fix for that.

src/commands/key2ds.rs
tertsdiepraam: I'm not sure I follow. I need the `dnskey` value later, so I think the current implementation makes sense? What are you suggesting exactly?

ximon18: Ah ok, ignore me.

src/commands/key2ds.rs
```diff
@@ -0,0 +1,169 @@
...
+            let rr = Record::new(owner, class, ttl, ds);
+
+            if self.write_to_stdout {
+                println!("{}", rr.display_zonefile(false));
+            } else {
+                let owner = owner.fmt_with_dot();
+                let sec_alg = sec_alg.to_int();
+                let filename =
+                    format!("K{owner}+{sec_alg:03}+{key_tag:05}.ds");
+                let mut out_file = File::create(&filename).map_err(|e| {
```
bal-e, line 128: Would it make sense to use `File::create_new()` here?

tertsdiepraam: That is, if we have consensus that we want to break compat here, I'd be happy to change it.

tertsdiepraam:
Little update: I changed the arguments for the dnst version so that we can use `--force/-f` for overwriting files and turn the ignoring sep option into the more descriptive `--ignore-sep`. I also added file opening/creation into `Env`, which can now be set up with a temp dir. Adds a bit of complexity, but not too much since we already have to use `Env` everywhere and the tests are quite nice if I do say so myself.

src/commands/key2ds.rs
mozzieongit: Now that we have different argument parsing for ldns and dnst, and we can use different defaults, we can probably also introduce differences here. Maybe a simple `running_as_ldns: bool` (clap-hidden) member is enough or should it be an enum like `MultiCallMode::{LDNS,DNST}`/`CompatMode::{None/DNST,LDNS}` (or similar)?

Cargo.toml
```diff
@@ -10,9 +10,16 @@ path = "src/bin/ldns.rs"
...
+domain = { version = "0.10.3", git = "https://github.com/NLnetLabs/domain.git", features = [
+    "zonefile",
+    "bytes",
+    "unstable-validate",
+] }
 lexopt = "0.3.0"
 
 # for implementation of nsec3 hash until domain has it stabilized
-octseq = { version = "0.5.1", features = ["std"] }
+octseq = { version = "0.5.2", features = ["std"] }
```
mozzieongit, line 21: Will this be an issue when merging the deletion of these lines with the nsec3 pr #6?

src/commands/key2ds.rs
```diff
@@ -0,0 +1,497 @@
...
+
+        let Some(keyfile) = keyfile else {
+            return Err("No keyfile given".into());
+        };
+
+        Ok(Self {
+            ignore_sep,
+            write_to_stdout,
+            algorithm,
+            // Preventing overwritten files is a dnst feature that is not
```
mozzieongit, line 116:
```suggestion
            // Preventing overwriting files is a dnst feature that is not
```

src/commands/key2ds.rs
mozzieongit: > Now that we have different argument parsing for ldns and dnst, and we can use different defaults, we can probably also introduce differences here. Maybe a simple `running_as_ldns: bool` (clap-hidden) member is enough or should it be an enum like `MultiCallMode::{LDNS,DNST}`/`CompatMode::{None/DNST,LDNS}` (or similar)?

Oh you implemented that already with the force_overwrite

tertsdiepraam: Yep!

src/env/fake.rs
```diff
@@ -53,6 +58,43 @@ impl Env for FakeEnv {
...
+        P: AsRef<Path>,
+    {
+        File::create(self.in_cwd(path))
+    }
+
+    fn file_create_new<P>(&self, path: P) -> Result<File, io::Error>
+    where
+        P: AsRef<Path>,
+    {
+        File::create_new(dbg!(self.in_cwd(path)))
```
mozzieongit, line 80: Is the `dbg!` meant to be there?

mozzieongit (commented):
Apart from the minor comments, it looks good to me. Also, very thorough testing.

src/env/fake.rs
tertsdiepraam: Definitely not haha

Cargo.toml
tertsdiepraam: Possibly 🤷🏻 I suppose we'll figure it out once we get there

PR #5, opened 2024-09-11
Add private key file parsing and generation

mozzieongit:
I've implemented parsing and generating the private key file format used to store the private key components for dnssec/tsig keys in the format used by bind and nsd.
The `...KeyData` structs are (potentially) temporary data stores for the needed key components, until it's clear where and how these functions will get integrated.
I tested the parsing and generation together by reading/parsing test key files (generated with ldns) and comparing the output of the generate function (with the data of the just parsed file) to the original key files.

Known quirks:
- The human readable algorithm "comment" in the private key files differs from ldns in the cases of `RSASHA1-NSEC3-SHA1`, `RSAMD5`, and `DSA-NSEC3-SHA1`, because ldns doesn't use the iana provided mnemonic/name but slightly different ones (`RSASHA1_NSEC3`, `RSA`, `DSA_NSEC3`)

In `gen_private_key_file_text()` I tried both the long `format!()` version and using a bunch of `String::push_str()`, but found this easier to read when formatted with `rustfmt`.
I didn't find a formal definition of the Private-key-format, so I created one in ABNF (comment in `parse_private_key_file()`) mostly for fun, but might also be something useful in the future? (Can of course be put somewhere else or removed entirely)

This code doesn't have to be put in dnst, but I wanted to create a PR so that we can track its existence.

Also feel free to nit pick my use of Rust.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -3,3 +3,4 @@ pub use self::args::Args;
 pub mod args;
 pub mod commands;
 pub mod error;
+pub mod utils;
--- a/src/utils/mod.rs
+++ b/src/utils/mod.rs
@@ -1,0 +1,1 @@
+pub mod private_key_file_parser;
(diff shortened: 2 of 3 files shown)


ximon18:
See PR #9 instead.

PR #6, opened 2024-10-15
Add nsec3-hash command.

ximon18:
Note: A minimal working version of this command already existed in `main` but used code copied from `domain` rather than depending on `domain`.

Currently depends on `domain` branch `initial-nsec3-generation`, see https://github.com/NLnetLabs/domain/pull/416.

Contains several kinds of tests:
- Unit tests.
- Integration tests in `tests/` (ignored by default) that compare outputs to a host installed `ldns-nsec3-hash` command.

Fuzz tests and man page changes were moved out to https://github.com/NLnetLabs/dnst/pull/24 and https://github.com/NLnetLabs/dnst/pull/25.


ldns-nsec3-hash
tertsdiepraam, line 1: I don't think this should be committed or at least not in the root.

We could make a folder with symlinks for testing or maybe some way to trigger the ldns parsing from `dnst`? Possibly with a separate binary target?

src/commands/nsec3hash.rs
```diff
@@ -1,55 +1,61 @@
...
 
 use super::{parse_os, parse_os_with, LdnsCommand};
 
 #[derive(Clone, Debug, clap::Args)]
+#[cfg_attr(feature = "arbitrary", derive(arbitrary::Arbitrary))]
 pub struct Nsec3Hash {
     /// The hashing algorithm to use
     #[arg(
-        long,
+        long = "algorithm",
```
tertsdiepraam, line 20: Agreed!

src/commands/nsec3hash.rs
```diff
@@ -1,55 +1,61 @@
...
 const LDNS_HELP: &str = "\
 ldns-nsec3-hash [OPTIONS] <domain name>
   prints the NSEC3 hash of the given domain name
-
-  -a <algorithm> hashing algorithm number
-  -t <number>    iterations
-  -s <string>    salt in hex\
+-a [algorithm] hashing algorithm
+-t [number] number of hash iterations
+-s [string] salt
```
tertsdiepraam, line 58: The indentation was intentional because I liked it better, but I guess for compat this would be more "correct". The angle brackets were "more correct" (for some definition of correct based on convention) because they indicate a required value where square brackets mean optional values.

tests/vs-ldns.rs
```diff
@@ -0,0 +1,101 @@
...
+        assert_ldns_eq_dnst(
+            "ldns-nsec3-hash",
+            &["-t", &iterations.to_string(), "example.com"],
+        );
+    }
+
+    assert_exit_code_ldns_eq_dnst("ldns-nsec3-hash", &[]);
+}
+
+fn create_temp_symlink(ldns_command: &str) -> std::path::PathBuf {
```
tertsdiepraam, line 23: Ha this is clever. It's also a bit complicated though, maybe we should just assert that the ldns parsing creates the correct structs without a separate process?

ldns-nsec3-hash
ximon18: This file should not have been committed.

tests/vs-ldns.rs
ximon18: I don't like it at all and now that I think about it is shouldn't be needed.

We shouldn't test this here using real sym links, we should just test by passing the right argv0 to the outermost layer of our code, otherwise we're just testing O/S behaviour...

And for a sanity check that the behaviour when invoked as a sym link works we can do that in a Ploutos package test script which also validates that the right sym links are created on package install.

I'll remove this sym linking from the PR.

ximon18: And as you said a way to force test "as if" ldns would be nice and as you suggested we can add that as a separate bin that only differs in main by not including argv0 in the args passed to the argument parsing code.

src/commands/nsec3hash.rs
ximon18: I'm reluctant to start improving the ldns-xxx commands at this stage, I'd rather get things implemented and basically working and as compatible as we can easily do. The temptation to start tweaking things if we start might be great and where do we stop? :-)

tests/vs-ldns.rs
```diff
@@ -0,0 +1,116 @@
...
+    );
+    assert!(
+        !cmd1_output.stdout.is_empty(),
+        "Unexpected stdout content for ldns command: {}: {:?}",
+        cmd1.join(" "),
+        std::str::from_utf8(&cmd1_output.stdout)
+    );
+    assert!(
+        !cmd2_output.stdout.is_empty(),
+        "Unexpected stdout content for dnst command: {}: {:?}",
```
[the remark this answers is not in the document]
ximon18: No, it says !empty.

doc/manual/source/conf.py
```diff
@@ -189,8 +189,8 @@
 # (source start file, name, description, authors, manual section).
 man_pages = [
     ('man/dnst', 'dnst', 'DNS Management Tools', author, 1),
-    ('man/dnst-nsec3-hash', 'dnst-nsec3-hash', 'DNS Management Tools', author,
-     1),
+    ('man/dnst-nsec3-hash', 'dnst-nsec3-hash', 'print out the NSEC3 hash for a domain name', author, 1),
+    ('man/ldns-nsec3-hash', 'ldns-nsec3-hash', 'print out the NSEC3 hash for a domain name', author, 1),
```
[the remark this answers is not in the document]
ximon18: No?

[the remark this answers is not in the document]
mozzieongit: oh, I misread the second command :facepalm:

tests/vs-ldns.rs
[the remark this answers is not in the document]
mozzieongit: Am I understanding assert wrong then?
stdout should be not empty aka should be something, therefore if it is empty, print the error message?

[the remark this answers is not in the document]
ximon18: Oh, maybe I'm wrong, I'll check.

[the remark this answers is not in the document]
ximon18: I think you are right.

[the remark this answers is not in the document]
ximon18: I decided that the tests were correct but the assert messages were inverted, and have corrected the messages.

src/commands/nsec3hash.rs
```diff
@@ -111,71 +118,275 @@ impl Nsec3Hash {
...
+            Self::num_to_nsec3_alg(num)
         } else {
             Nsec3HashAlg::from_mnemonic(arg.as_bytes()).ok_or("unknown algorithm mnemonic")
         }
     }
+
+    pub fn parse_nsec3_alg_as_num(arg: &str) -> Result<Nsec3HashAlg, &'static str> {
+        match arg.parse() {
+            Ok(num) => Self::num_to_nsec3_alg(num),
+            Err(_) => Err("malformed algorith number"),
```
tertsdiepraam, line 142:
```suggestion
            Err(_) => Err("malformed algorithm number"),
```

src/lib.rs
```diff
@@ -1,29 +1,37 @@
...
+    })
+}
 
-pub fn try_ldns_compatibility<I: IntoIterator<Item = OsString>>(args: I) -> Option<Args> {
+pub fn try_ldns_compatibility<I: IntoIterator<Item = OsString>>(args: I) -> Result<Args, Error> {
     let mut args_iter = args.into_iter();
-    let binary_path = args_iter.next()?;
+    let binary_path = args_iter
+        .next()
+        .ok_or::<Error>("Missing binary name".into())?;
```
tertsdiepraam, line 25: I think the type annotation is only necessary because the `?` adds an `into` as well, so removing the `.into()` might clean this up a bit.

tests/vs-ldns.rs
```diff
@@ -0,0 +1,116 @@
...
+const DNST_NSEC3_SUBCMD: &str = "nsec3-hash";
+
+#[ignore = "should only be run if ldns command line tools are installed"]
+#[test]
+fn nsec3_hash() {
+    // Note: ldns-nsec3-hash defaults NSEC3 iterations to 1, while dnst
+    // nsec-hash defaults NSEC3 iterations to 0.
+    assert_cmds_eq(
+        &[LDNS_NSEC3_CMD, TEST_ZONE_NAME],
+        &[DNST_NSEC3_SUBCMD, "--iterations", "1", TEST_ZONE_NAME],
```
tertsdiepraam, line 14: I'm not sure about how useful these tests are when the defaults are different? It would make more sense to run this with the ldns compatibility mode I suppose.

src/lib.rs
ximon18: The latest code changes this behaviour, but I noted there still seems to be unwanted error output in some cases, e.g. on `--help`, so I would rather just merge your env PR to main and merge that into this and see what we end up with.

tests/vs-ldns.rs
ximon18: I think both make sense, otherwise we don't have a complete sanity check of the `dnst` command as a black box (though we do have that in the packaging workflow sanity check).

mozzieongit (dismissed):
I've not tested the new changes, but LGTM

src/lib.rs
```diff
@@ -1,29 +1,37 @@
...
 
-pub mod args;
-pub mod commands;
-pub mod error;
+pub fn parse_args<I: IntoIterator<Item = OsString>, T: Fn() -> I>(
+    args_provider: T,
+) -> Result<Args, Error> {
+    try_ldns_compatibility(args_provider()).or_else(|_| {
+        Args::try_parse_from(args_provider()).map_err(|err| Error::new(err.to_string().as_str()))
+    })
```
tertsdiepraam, line 18: This is subtly wrong. The `try_ldns_compatibility` function has three outcomes:
1. The command was not matched
2. The command matched and parsing succeeded
3. The command matched and parsing failed

You've now combined the first and third case even though they should be handled differently.

I think you're also removing the color from the clap output by turning it into a string.

src/commands/nsec3hash.rs
```diff
@@ -111,71 +117,274 @@ impl Nsec3Hash {
...
+            assert!(parse_cmd_line(&["-s"]).is_err());
+            assert!(parse_cmd_line(&["-s", "nlnetlabs.nl"]).is_err());
+            assert!(parse_cmd_line(&["-s", "NOTHEX", "nlnetlabs.nl"]).is_err());
+            assert!(parse_cmd_line(&["-s", &"aa".repeat(256), "nlnetlabs.nl"]).is_err());
+        }
+
+        //------------ Helper functions ------------------------------------------
+
+        fn parse_cmd_line(args: &[&str]) -> Result<Args, Error> {
+            parse_args(|| ["ldns-nsec3-hash"].iter().chain(args).map(Into::into))
```
tertsdiepraam, line 380: This is what this would look like after #23:
```rust
parse_args(FakeEnv {
    args: ["ldns-nsec3-hash"].iter().chain(args).map(Into::into)).collect(),
    ..Default::default(),
})
```

src/commands/nsec3hash.rs
```diff
@@ -111,71 +117,274 @@ impl Nsec3Hash {
...
+        }
+
+        fn assert_cmd_eq(args: &[&str], expected_output: &str) {
+            let parsed_args = parse_cmd_line(args).unwrap();
+            let mut captured_stdout = vec![];
+            parsed_args.execute(&mut captured_stdout).unwrap();
+            assert_eq!(str::from_utf8(&captured_stdout), Ok(expected_output));
```
tertsdiepraam, line 387: This is what this would look like after #23:
```rust
let env = FakeEnv {
    args: ["ldns-nsec3-hash"].iter().chain(args).map(Into::into)).collect(),
    ..Default::default(),
}

let res = run(&env);
assert_eq!(res, 0);
assert_eq!(env.get_stdout(), expected_output);
assert_eq!(env.get_stderr(), expected_output);
```

Cargo.toml
```diff
@@ -9,10 +9,10 @@ name = "ldns"
 path = "src/bin/ldns.rs"
 
 [dependencies]
-clap      = { version = "4", features = ["derive"] }
-#domain   = "0.10.3"
-domain    = { git = "https://github.com/NLnetLabs/domain/", branch = "initial-nsec3-generation", features = ["ring", "unstable-validator"] }
-lexopt    = "0.3.0"
+clap = { version = "4.3.4", features = ["derive"] }
+domain = { version = "0.10.3", git = "https://github.com/NLnetLabs/domain.git", branch = "initial-nsec3-generation", features = ["unstable-validator", "zonefile"] }
```
ximon18, line 13: Why do you need feature "zonefile" ?

src/bin/ldns.rs
```diff
@@ -6,21 +6,21 @@
...
 use dnst::try_ldns_compatibility;
 
 fn main() -> ExitCode {
-    match run() {
+    let env = dnst::env::RealEnv;
+
+    let mut args = std::env::args_os();
+    args.next().unwrap();
+    let args =
+        try_ldns_compatibility(args).map(|args| args.expect("ldns commmand is not recognized"));
```
ximon18, line 17: Using `expect` here will presumably result in a rather ugly end user experience? Also, the LDNS command matching doesn't seem to be working, e.g.:

```
$ cargo build --release

$ ln -s target/release/dnst ldns-nsec3-blah

$ ./ldns-nsec3-blah 
Usage: ldns-nsec3-blah <COMMAND>

Commands:
  nsec3-hash  Prints the NSEC3 hash of a given domain name
  key2ds      Generate a DS RR from the DNSKEYS in keyfile
  help        Show the manual pages

Options:
  -h, --help     Print help
  -V, --version  Print version

File: Cargo.toml (hunk 12 above)
Comment (mozzieongit) [reply Cargo.toml:13]: For key2ds

File: src/bin/ldns.rs (hunk 13 above)
Comment (mozzieongit) [reply src/bin/ldns.rs:17]: The ldns binary is for testing only (see comment at the top of the file)

Comment (mozzieongit) [reply src/bin/ldns.rs:17]: The second part, probably needs looking into

File: src/commands/mod.rs [hunk 14]
```diff
@@ -34,9 +33,10 @@ pub enum Command {
 }
 
 impl Command {
-    pub fn execute<W: Write>(self, writer: &mut W) -> Result<(), Error> {
+    pub fn execute(self, env: impl Env) -> Result<(), Error> {
         match self {
-            Self::Nsec3Hash(nsec3hash) => nsec3hash.execute(writer),
+            Self::Nsec3Hash(nsec3hash) => nsec3hash.execute(env),
+            Self::Key2ds(key2ds) => key2ds.execute(env),
```
ximon18, line 39: Why does this PR for NSEC3 hashing add a mapping for `key2ds`?

mozzieongit: Looks like the merge didn't get recognized by github

src/commands/nsec3hash.rs
```diff
@@ -152,14 +152,16 @@ impl Nsec3Hash {
...
         let hash =
             nsec3_hash::<_, _, Vec<u8>>(&self.name, self.algorithm, self.iterations, &self.salt)
                 .map_err(|err| format!("Error creating NSEC3 hash: {err}"))?
                 .to_string()
                 .to_lowercase();
-        writeln!(writer, "{}.", hash)
-            .map_err(|err| Error::from(format!("Error writing to output: {err}")))
+
+        let mut out = env.stdout();
+        writeln!(out, "{}.", hash);
```
ximon18, line 163:
```suggestion
        writeln!(env.stdout() "{}.", hash);
```

src/commands/nsec3hash.rs
```diff
@@ -184,6 +187,12 @@ mod tests {
         // nsec3-hash`` or as `ldns-nsec3-hash`.
         #[test]
         fn execute() {
+            let env = FakeEnv {
+                cmd: FakeCmd::new(["doesn't matter"]),
```
ximon18, line 191:
```suggestion
                cmd: FakeCmd::new(["unused"]),
```

src/commands/mod.rs
mozzieongit: Seems like this is only shown as a change if showing the recent changes/commits (e.g. since last review) and not when viewing all commits.

src/lib.rs
```diff
@@ -1,38 +1,54 @@
...
         .to_str()
-        .ok_or("Binary file name is not valid Unicode")?;
+        .ok_or("Binary file name is not valid unicode")?;
 
-    match binary_name {
+    let res = match binary_name {
+        "ldns-key2ds" => Key2ds::parse_ldns_args(args_iter),
         "ldns-nsec3-hash" => Nsec3Hash::parse_ldns_args(args_iter),
-        _ => Err("Unrecognised LDNS command".into()),
+        _ => return Ok(None),
```
ximon18, line 31: This should presumably be `Err` rather than `Ok(None)`.

src/bin/ldns.rs
tertsdiepraam: That's expected in my opinion. The idea was to support the multicall behaviour for the ldns-tools, but anything else would be `dnst`. Is your suggestion to make it dependent on the `ldns-` prefix?

src/lib.rs
tertsdiepraam: I was using:
- `Ok(Some(args))`: correctly parsed ldns command
- `Ok(None)`: no ldns command matched
- `Err(err)`: an ldns command failed.

If we return an error, we get into trouble because we need to recover (i.e. call clap parsing). We could do `Option<Result<_, _>>` instead of `Result<Option<_>, _>` but that will just lead to `.transpose()?` somewhere else.

src/bin/ldns.rs
tertsdiepraam: And yes, the expect is because it's just an internal tool :)

src/lib.rs
tertsdiepraam: See: https://github.com/NLnetLabs/dnst/pull/33

src/bin/ldns.rs
tertsdiepraam: See: https://github.com/NLnetLabs/dnst/pull/33

ximon18:
LGTM @tertsdiepraam.

PR #7, opened 2024-10-16
Notify

tertsdiepraam:
This PR includes:
- The implementation of `ldns-notify` and `dnst notify`.
- Stelline in the env.
- Printing version and help from ldns argument parsing.

### DIFF
--- a/src/commands/mod.rs
+++ b/src/commands/mod.rs
@@ -109,7 +109,7 @@ pub trait LdnsCommand {
     fn parse_ldns_args<I: IntoIterator<Item = OsString>>(args: I) -> Result<Args, Error> {
         match Self::parse_ldns(args) {
             Ok(c) => Ok(c),
-            Err(e) => Err(format!("Error: {e}\n\n{}", Self::HELP).into()),
+            Err(e) => Err(format!("{e}\n\n{}", Self::HELP).into()),
         }
     }
 
--- a/src/commands/notify.rs
+++ b/src/commands/notify.rs
@@ -122,7 +122,7 @@ Ldns notify utility
 
  Supported options:
         -z zone         The zone
-        -I <address>    source address to query from
+        -I <address>    source address to query from (currently unsupported)
         -s version      SOA version number to include
         -y <name:key[:algo]>    specify named base64 tsig key, and optional an
                         algorithm (defaults to hmac-md5.sig-alg.reg.int)
@@ -132,7 +132,7 @@ Ldns notify utility
         -r num          max number of retries (15)
         -h              Print this help information
 
-Report bugs to <dns-team@nlnetlabs.nl>\
+Report bugs to <dns-team@nlnetlabs.nl>
 ";
 
 impl LdnsCommand for Notify {
@@ -157,7 +157,7 @@ impl LdnsCommand for Notify {
                     let val = parser.value()?;
                     zone = Some(parse_os("zone (-z)", &val)?);
                 }
-                Arg::Short('I') => todo!(),
+                Arg::Short('I') => return Err("The -I option is currently unsupported".into()),
                 Arg::Short('s') => {
                     let val = parser.value()?;
                     soa_version = Some(parse_os("soa version (-s)", &val)?);


src/commands/notify.rs
```diff
@@ -0,0 +1,287 @@
...
+        let mut config = dgram::Config::new();
+        config.set_max_retries(self.retries as u8);
+        let connection = dgram::Connection::with_config(UdpConnect::new(socket), config);
+        let connection = tsig::Connection::new(key, connection);
+
+        let req = RequestMessage::new(msg).unwrap();
+        let mut req = connection.send_request(req);
+    
+        let when = Local::now();
+        let msg = req.get_response().await.map_err(|e| format!("warning: could not send message to {socket}: {e}"))?;
```
mozzieongit, line 265: I would rephrase this error, as I think it's misleading. The request was sent (at least in my test), but the error comes from validating the response from the server. So maybe more like: `warning: erroneous response from {socket}: {e}`. I came across this when using tsig, but the server didn't use tsig in the response.

mozzieongit: Or will there be different errors? Some from failed sending and some from failed receiving?

mozzieongit (changes requested):
It looked like you're mostly done, so I checked it out. There's a left-over TODO though.

src/commands/notify.rs
```diff
@@ -0,0 +1,287 @@
...
+    #[arg(short = 'd', long = "debug")]
+    debug: bool,
+
+    /// Max number of retries
+    #[arg(short = 'r', long = "retries", default_value = "15")]
+    retries: usize,
+
+    // Hidden extra argument for `-?` to trigger help, which ldns supports.
+    #[arg(short = '?', action = ArgAction::Help, hide = true)]
+    compatible_help: (),
```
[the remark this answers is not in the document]
tertsdiepraam: Don't know where I got that from :smile:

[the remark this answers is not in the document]
tertsdiepraam: Ah this is where I got it from:

https://github.com/NLnetLabs/ldns/blob/b87b8e1d212b43f225891570ebc5045c6331dd8a/examples/ldns-notify.c#L282

But it's not part of the string here:

https://github.com/NLnetLabs/ldns/blob/b87b8e1d212b43f225891570ebc5045c6331dd8a/examples/ldns-notify.c#L193

So it's not supported indeed.

src/commands/notify.rs
```diff
@@ -0,0 +1,287 @@
...
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        // TODO: better error messages
+        let Some((mut name, rest)) = s.split_once(':') else {
+            return Err("invalid tsig string".into());
+        };
+
+        let mut key;
+        let mut algorithm;
+        if let Some((k, a)) = s.split_once(':') {
```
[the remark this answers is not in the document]
tertsdiepraam: Oof that's sloppy 😅

src/commands/mod.rs
```diff
@@ -11,6 +12,14 @@ pub enum Command {
     #[command(name = "nsec3-hash")]
     Nsec3Hash(self::nsec3hash::Nsec3Hash),
 
+    /// Send a NOTIFY packet to DNS servers
+    ///
+    /// This tells them that an updated zone is available at the master servers. It can perform TSIG
```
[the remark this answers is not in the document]
tertsdiepraam: Is this just newer terminology or does it mean anything different? In the first case, I definitely agree. But this came from the manpage of ldns-notify, so decided to copy that first.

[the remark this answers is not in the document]
mozzieongit: It's quite old terminology actually (which had been replaced by master/slave some decades ago), but is now changed to primary/seconday again, as in many other environments (git, parallel prog., etc.).
It means the same, and is known terminology, the ldns man pages are just old.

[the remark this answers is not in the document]
ximon18: I think our preference is to avoid using words that are known to be offensive to some and use alternate replacement terms where available. Master and slave should indeed I believe be replaced by primary and secondary. [RFC 9499](https://datatracker.ietf.org/doc/html/rfc9499) states:

> Slave server: See "Secondary server".
> Master server: See "Primary server".

src/commands/notify.rs
tertsdiepraam: Both are possible yeah and unfortunately not clearly distinguished. I'll update the text to mention both possibilities.

src/commands/mod.rs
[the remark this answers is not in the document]
tertsdiepraam: Excellent! Then I'll do that, because that terminology should indeed disappear

src/commands/notify.rs
```diff
@@ -0,0 +1,324 @@
...
+        let connection = tsig::Connection::new(key, connection);
+
+        let req = RequestMessage::new(msg).unwrap();
+        let mut req = connection.send_request(req);
+
+        let when = Local::now();
+        let msg = req
+            .get_response()
+            .await
+            .map_err(|e| format!("warning: could not send message to {socket}: {e}"))?;
```
mozzieongit, line 290: Here as well?
```suggestion
            .map_err(|e| format!("warning: reply was not received or erroneous from {socket}: {e}"))?;
```

src/commands/notify.rs
```diff
@@ -0,0 +1,340 @@
...
+    }
+}
+
+impl FromStr for TSigInfo {
+    type Err = Error;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        // TODO: better error messages
+        let Some((mut name, rest)) = s.split_once(':') else {
+            return Err("invalid tsig string".into());
```
mozzieongit, line 61: Maybe "invalid TSIG key format" or is the TODO just forgotten?

mozzieongit:
FYI: ldns-notify supports `-v`

src/commands/notify.rs
```diff
@@ -0,0 +1,710 @@
...
+    // #[arg(short = 'I', required = false)]
+    // source_address: (),
+    //
+    /// SOA version number to include
+    #[arg(short = 's')]
+    soa_version: Option<u32>,
+
+    /// A base64 tsig key and optional algorithm to include
+    #[arg(
+        short = 'y',
```
bal-e, line 124: Why is this `-y`?  Do we want to change it to something more obvious, or should we keep it as-is for ldns compatibility?

src/commands/notify.rs
```diff
@@ -0,0 +1,710 @@
...
+            for socket in hosts.port_iter(self.port) {
+                self.notify_host(&env, socket, msg.clone(), server, &tsig)
+                    .await;
+            }
+        }
+
+        Ok(())
+    }
+
+    async fn notify_host(
```
bal-e, line 351: Could you document what this and the other internal functions do?

src/commands/notify.rs
```diff
@@ -0,0 +1,710 @@
...
+
+    async fn notify_host_with_tsig(
+        &self,
+        env: impl Env,
+        socket: SocketAddr,
+        msg: Message<Vec<u8>>,
+        key: Key,
+    ) -> Result<Response<Bytes>, Error> {
+        let mut config = dgram::Config::new();
+        config.set_max_retries(self.retries as u8);
```
bal-e, line 386: Maybe the `retries` field should just be a `u8`?

src/commands/notify.rs
```diff
@@ -0,0 +1,710 @@
...
+        let connection = dgram::Connection::with_config(env.dgram(socket), config);
+        let connection = tsig::Connection::new(key, connection);
```
bal-e, line 388: You can combine this with the `without_tsig` method by using `Box<dyn>`.

bal-e (commented):
Really well done, @tertsdiepraam!  This is some complex code but I like the way you've managed to get through it!

src/commands/notify.rs
ximon18: FYI other tools also use `-y`, e.g. `dig`.

src/commands/notify.rs
```diff
@@ -0,0 +1,710 @@
...
+            // This is different from ldns's default of
+            // hmac-md5.sig-alg.reg.int but we don't support that algorithm.
+            algorithm = "hmac-sha512";
+        }
+
+        // With dig TSIG keys are also specified with -y,
+        // but our format is: <name:key[:algo]>
+        //      and dig's is: [hmac:]name:key
+        //
+        // When we detect an unknown tsig algorithm in algo,
```
ximon18, line 80:
```suggestion
        // When we detect an unknown TSIG algorithm in algo,
```

src/commands/notify.rs
tertsdiepraam: There's not a lot to document, I think. I added something, but I'm not sure what you'd like to see. What should I write?

bal-e: Even a one-line note is enough.  As I'm reading code, I like having a quick summary idea of what each high-level function does ... and I couldn't figure out one for this function, at least not at a glance.  Something simple like "Send a notify packet to a specific server"?

bal-e (approved):
Nicely done, @tertsdiepraam!

src/commands/notify.rs
```diff
@@ -142,20 +109,111 @@ pub struct Notify {
...
     servers: Vec<String>,
 }
 
+const LDNS_HELP: &str = "\
+usage: ldns-notify [other options] -z zone <servers>
+Ldns notify utility
+
+ Supported options:
+        -z zone         The zone
+        -I <address>    source address to query from
```
mozzieongit, line 125: Maybe we should mention that this is currently unsupported?

src/commands/notify.rs
```diff
@@ -142,20 +109,111 @@ pub struct Notify {
...
+
+        let mut parser = lexopt::Parser::from_args(args);
+
+        while let Some(arg) = parser.next()? {
+            match arg {
+                Arg::Short('z') => {
+                    let val = parser.value()?;
+                    zone = Some(parse_os("zone (-z)", &val)?);
+                }
+                Arg::Short('I') => todo!(),
```
mozzieongit, line 160: Maybe this should be an error instead saying it's currently unsupported?

mozzieongit (changes requested):
Last complaints, I swear! _... maybe_ :D

src/commands/notify.rs
tertsdiepraam: You're absolutely right of course!

src/commands/notify.rs
tertsdiepraam: Good idea!

tertsdiepraam:
@mozzieongit do your worst! Give me that great feedback! 😄

PR #8, opened 2024-10-17
Add signzone command.

ximon18:
Currently depends on `domain` branch `initial-nsec3-generation`, which has had multiple branches/PRs merged into it (see https://github.com/NLnetLabs/domain/pull/416)

Supports:
- The basic command line arguments `zonefile key [key [key]]` and the NSEC3 arguments `-n`, `-a`, `-t`, `-s` and `-p`.
- Additional command line arguments `-o`, `-i`, `d`, `-e`, `-f`,`-u`, `-A`, `-U` and `-v`.
- `-z` and `-Z` for ZONEMD

Partially supports:
- Command line argument `-b` (support for [Bubble Babble](https://bohwaz.net/archives/web/Bubble_Babble.html) DS comment output is not planned at present).

Lacks but should have support for:
- Support for parsing some record types. These require support to be added to the underlying `domain` library. See:
  - SSHFP: https://github.com/NLnetLabs/domain/issues/451
  - TLSA: https://github.com/NLnetLabs/domain/issues/452
  - OPENPGPKEY: https://github.com/NLnetLabs/domain/issues/453
  - IPSECKEY: https://github.com/NLnetLabs/domain/issues/454
  - RFC 3597 unknown record type with zero length data: https://github.com/NLnetLabs/domain/pull/475

Lacks and do not plan to add support for:
- OpenSSL engine related arguments.
- [Bubble Babble](https://bohwaz.net/archives/web/Bubble_Babble.html) DS comment output.

Other:
- Partial signing and re-signing: LDNS has strange behaviour here, so `dnst` removes DNSSEC records on loading already signed zonefiles.
- Verify that it is expected that the `-U` option causes a warning from `dnssec-verify` (it also does so for the original `ldns-signzone` when using `-U` so I think this is fine). We should think do we want to support the `-U` option for `dnst signzone`?
- Rendering zonefile entries may not exactly match the output of `ldns-signzone` as the behaviour is determined by the `domain` crate. (see `ldns_rr2buffer_str_fmt()` in LDNS). Known differences are:
  - Some RDATA values are cased differently, but all known examples relate to RFCs that say that the text in question is "case-insensitive", so this is a difference but not an error.

Additional related DRAFT PRs:
- https://github.com/NLnetLabs/dnst/pull/35
  - Requires: https://github.com/NLnetLabs/domain/pull/448

This PR adds automated tests but has also been tested manually against the original `ldns-signzone` and `dnssec-signzone`.

### DIFF
--- a/doc/manual/source/man/dnst-signzone.rst
+++ b/doc/manual/source/man/dnst-signzone.rst
@@ -21,7 +21,7 @@ Arguments
 .. option:: <ZONEFILE>
 
       The zonefile to sign. Any existing NSEC(3) and/or RRSIG resource records
-      will be skipped when loaded the file.
+      will be skipped when loading the file.
 
 .. option:: <KEY>...
 


Cargo.toml
```diff
@@ -4,9 +4,21 @@ version = "0.1.0"
...
+    "unstable-zonetree",
+    "zonefile",
+] }
 
 # for implementation of nsec3 hash until domain has it stabilized
 octseq = { version = "0.5.1", features = ["std"] }
 ring = { version = "0.17" }
+
+# for the signzone command
+openssl = { version = "0.10.57" } # required for RSASHA1 and 1024 bit key support (as ring doesn't support these)
```
bal-e, line 24: Won't `domain` include this for us?  Or should we be exporting it under `domain::dep`?

src/commands/signzone.rs
```diff
@@ -0,0 +1,285 @@
...
+                        "Unable to parse BIND formatted private key file '{}': {}",
+                        private_key_path, err
+                    ))
+                })?;
+
+            let public_key_info: domain::validate::Key<Bytes> =
+                domain::validate::Key::parse_from_bind(&public_data).map_err(|err| {
+                    Error::from(format!(
+                        "Unable to parse BIND formatted public key file '{}': {}",
+                        private_key_path, err
```
bal-e, line 150: Wrong path being printed here.

src/commands/signzone.rs
```diff
@@ -0,0 +1,285 @@
...
+            let secret_key =
+                domain::sign::SecretKeyBytes::parse_from_bind(&private_data).map_err(|err| {
+                    Error::from(format!(
+                        "Unable to parse BIND formatted private key file '{}': {}",
+                        private_key_path, err
+                    ))
+                })?;
+
+            let public_key_info: domain::validate::Key<Bytes> =
+                domain::validate::Key::parse_from_bind(&public_data).map_err(|err| {
```
bal-e, line 147: It's unfortunate that this is generic over `Octs` right now, but I'd suggest omitting the explicit type for the variable and just writing `domain::validate::Key::<Bytes>::parse_from_bind(...)`, that'll be a bit shorter.

src/commands/signzone.rs
```diff
@@ -0,0 +1,285 @@
...
+        let mut records = self.load_zone()?;
+
+        //---
+
+        // Import the specified key.
+        let mut keys = vec![];
+        for key_path in self.key_paths {
+            let old_ext = key_path.extension().unwrap().to_str().unwrap();
+            let new_ext = format!("{}.private", old_ext);
+            let private_key_path = key_path.with_extension(new_ext).display().to_string();
```
bal-e, line 120: This is quite a convoluted way to append `.private` / `.key` to the path.  While it doesn't look like `PathBuf` has any good methods, you can do `key_path.clone().as_mut_os_string().push(".private")`.

bal-e, line 120: I really dislike using `.display().to_string()` as the path you pass into fs operations.  Once you've appended the extension, use the `PathBuf` as is, and when you need to print it for errors, just call `.display()` there.

src/commands/signzone.rs
```diff
@@ -0,0 +1,285 @@
...
+            let private_data = std::fs::read_to_string(&private_key_path).map_err(|err| {
+                Error::from(format!(
```
bal-e, line 122: I think you can omit `Error::from`, the `?` operator can do that for you.

src/commands/signzone.rs
```diff
@@ -0,0 +1,285 @@
...
+            )
+            .map_err(|err| {
+                Error::from(format!(
+                    "Unable to import private key from file '{}': {}",
+                    private_key_path, err
+                ))
+            })?;
+
+            let signing_key = domain::sign::SigningKey::new(
+                public_key_info.owner().to_owned(),
```
bal-e, line 166: Prefer `.clone()` over `.to_owned()`, I think.

src/commands/signzone.rs
```diff
@@ -0,0 +1,285 @@
...
+                public_key_info.flags(),
+                key_pair,
+            );
+
+            keys.push(signing_key);
```
bal-e, line 171: Perhaps we want to move the above loading logic into a separate function here, and then just call `key_paths.into_iter().map(...).try_collect()?`.

src/commands/signzone.rs
```diff
@@ -0,0 +1,285 @@
...
+            records.extend(recs.into_iter().map(Record::from_record));
+            records.insert(Record::from_record(param)).unwrap();
+            hashes
+        } else {
+            let nsecs = records.nsecs::<Bytes>(&apex, ttl);
+            records.extend(nsecs.into_iter().map(Record::from_record));
+            None
+        };
+
+        if !self.hash_only {
```
bal-e, line 204: It isn't immediately clear what "not only hashing" means, could you add a comment saying that this block will actually insert signatures for the records?

src/commands/signzone.rs
```diff
@@ -0,0 +1,285 @@
...
+                .unwrap();
+        } else {
+            records.write(&mut std::io::stdout().lock()).unwrap();
+        }
+
+        Ok(())
+    }
+
+    fn load_zone(&self) -> Result<SortedRecords<StoredName, StoredRecordData>, Error> {
+        let mut zone_file = File::open(self.zonefile_path.as_path())?;
```
bal-e, line 231: Prefer `&self.zonefile_path`.

src/commands/signzone.rs
```diff
@@ -0,0 +1,285 @@
...
+        let mut reader = inplace::Zonefile::load(&mut zone_file).unwrap();
+        if let Some(origin) = &self.origin {
+            reader.set_origin(origin.to_owned());
```
bal-e, line 234: Instead of calling `.to_owned()` within the function, just write `if let Some(origin) = self.origin.clone()`.

src/commands/signzone.rs
```diff
@@ -0,0 +1,285 @@
...
+        }
+        let mut records = SortedRecords::new();
+        for entry in reader {
+            let Ok(entry) = entry else {
+                return Err(Error::from(format!(
+                    "Invalid zone file: {}",
+                    entry.unwrap_err()
+                )));
+            };
```
bal-e, line 243: Prefer `let entry = entry.map_err(|err| format!(...))?;`.

src/commands/signzone.rs
```diff
@@ -0,0 +1,285 @@
...
+            match entry {
+                Entry::Record(record) => {
+                    let record: StoredRecord = record.flatten_into();
+                    records
+                        .insert(record)
+                        .expect("Invalid zone file: Duplicate record detected: {record:?}");
```
bal-e, line 249: An error that the user could cause shouldn't result in a panic.  Besides, I don't think the format string here will actually be interpolated.  Prefer `.map_err` and the `?` operator.

src/commands/signzone.rs
```diff
@@ -0,0 +1,285 @@
...
+                }
+                Entry::Include { .. } => {
+                    return Err(Error::from(
+                        "Invald zone file: $INCLUDE directive is not supported",
```
bal-e, line 253: Typo: "Invald"

src/commands/signzone.rs
```diff
@@ -0,0 +1,285 @@
...
+    fn find_apex(
+        records: &SortedRecords<StoredName, StoredRecordData>,
+    ) -> Result<(FamilyName<Name<Bytes>>, Ttl), std::io::Error> {
+        let soa = match records.find_soa() {
+            Some(soa) => soa,
+            None => {
+                return Err(std::io::Error::new(
+                    std::io::ErrorKind::Other,
+                    "cannot find SOA record",
+                ))
```
bal-e, line 270: Wouldn't it be better to return the crate-local `Error` type here?

bal-e (changes requested):
Good work, @ximon18!  We'll have to see whether the argument parsing needs to be changed to separate ldns / dnst, but the code generally looks good.

src/commands/signzone.rs
ximon18: That won't work because push doesn't return self.

ximon18: And unfortunately resolving that still leaves the problem that the resulting type doesn't impl `Display` due to `use of unstable library feature 'os_str_display'.

ximon18: I did a thing, I'm sure it's possible to do better. Take a look and see what you think.

src/commands/signzone.rs
ximon18: Nice.

src/commands/signzone.rs
ximon18: Fixed.

src/commands/signzone.rs
ximon18: Until now I hadn't done that because the key management code was still evolving and I didn't want to clean up too much until that stabilized. But that's merged now so yes I can extract this. The same btw goes for the hashing/signing code, I'm happy to leave that verbose/messy while the underlying library code is still changing.

src/commands/signzone.rs
ximon18: I'll introduce some sort of self-describing mode enum and use that instead.

src/commands/signzone.rs
ximon18: `try_collect()` appears to be unstable.

ximon18: I factored out the logic to a helper function but didn't use a functional iterator style collect-to-vec due to not being able to use the ? operator in a closure. Instead I used a simple for loop.

src/commands/signzone.rs
ximon18: Nice.

src/commands/signzone.rs
ximon18: Fair point.

mozzieongit:
dnst-signzone currently allows not specifying a key to sign with. In that case it just copies the original zone over into the .signed file (with no signatures ofc).

src/commands/signzone.rs
```diff
@@ -0,0 +1,709 @@
...
+
+        let mut writer = if out_file.as_os_str() == "-" {
+            // Box::new(env.stdout()) as Box<dyn Write>
+            // FIXME: env.stdout() uses impl fmt::Write, but because of
+            // domain::sign::records::SortedRecords::write_with_comments()
+            // we need io::Write here.
+            todo!()
+        } else {
+            Box::new(File::create(env.in_cwd(&out_file))?) as Box<dyn Write>
+        };
```
mozzieongit, line 406: Just so that it is written down: The merge of the env things introduced an issue in the current signzone code. Further down at L467 it uses writer with
```rust
records.write_with_comments(&mut writer, move |r, writer| match r.data() { ...
```
where it needs to implement `io::Write` currently, but `Env::stdout` uses `fmt::Write`. 
The idea mentioned earlier is to change the functions in `domain` to also use `fmt::Write`, as we don't need to be able to write arbitrary (non-UTF8) bytes.

ximon18: I updated the domain branch used by this PR to use `fmt::Write` instead of `io::Write`. The code in domain is in `src/sign/records.rs` which was in released code but never compiled as it had no `mod records.rs;` statement in `src/sign/mod.rs`. There is a `src/sign/bin/signzone.rs` orphaned example that might be affected so we should still check that. I have NOT updated this PR to use the new code in the domain branch.

ximon18: By the way, it might make more sense to:
- In the short term, move those functions out of domain into dnst.
- In the long term, add support to domain for some way of pretty printing zone entries with comments.
@tertsdiepraam: What are your thoughts on this?

ximon18:
> > dnst-signzone currently allows not specifying a key to sign with. In that case it just copies the original zone over into the .signed file (with no signatures ofc).
> 
> That's a bug.

Fixed in 907a634.

src/commands/signzone.rs
```diff
@@ -0,0 +1,1397 @@
...
+            zonemd,
+            allow_zonemd_without_signing,
+            use_nsec3,
+            algorithm,
+            iterations,
+            salt,
+            nsec3_opt_out_flags_only: true,
+            // TODO: is mixed up with flags_only? The commend on clap args says
+            // nsec3_opt_out is not originally supported, but flags_only is
+            nsec3_opt_out,
```
mozzieongit, line 382: The comment on the clap argument says that the `nsec3_opt_out` is not originally supported by ldns-signzone, but `nsec3_opt_out_flags_only` is. Is this a mix up in the ldns parsing then?

src/commands/signzone.rs
```diff
@@ -383,7 +383,7 @@ impl SignZone {
         res.map_err(|err| Error::from(format!("Invalid timestamp: {err}")))
     }
 
-    pub fn execute(self, env: impl Env) -> Result<(), Error> {
+    pub fn execute(self, env: impl Env, is_ldns: bool) -> Result<(), Error> {
```
tertsdiepraam, line 437: An alternative to this is to add a field to `Signzone` with `is_ldns` which is set to `false` by clap and `true` at the end of `parse_ldns`. That seems about as convenient, but also doesn't affect the other utils.

bal-e: I think developing better interfaces and functionality in the `dnst` versions of the commands we have will eventually necessitate this kind of parameter in more places.  We don't need to do that immediately, but I think it's something to keep in mind.

ximon18: I was aware of your idea to do it the other way @tertsdiepraam, but as I said I prefer it this way as I think it should be implemented once in a common single mechanism for all commands, not separately per command with potentially divergent implementations to solve a common problem.

tertsdiepraam: Sure! I made my case and stand by it, but it's alright! You've got a very practical solution here.

ximon18: I can live with going with the majority opinion in the team, maybe others don't agree with my stated reasons and prefer your approach. It's easier to choose among actual implementations than hypothetical designs so now we have both and can pick one.

src/commands/signzone.rs
```diff
@@ -0,0 +1,1782 @@
...
+    }
+}
+
+//------------ Commented -----------------------------------------------------
+
+/// Support for RTYPE specific zonefile comment generation.
+///
+/// Intended to be used to enable behaviour to be matched to that of the LDNS
+/// `ldns_rr2buffer_str_fmt()` function.
+trait Commented<T> {
```
ximon18, line 1336: Hmm, I just noticed that we have comment support in `ZonefileFmt` in `domain`. That should be used instead of this I guess.

ximon18:
In theory this PR needs `domain` branch `multiple-key-signing` (from https://github.com/NLnetLabs/domain/pull/418), however it currently (unintentionally) depends on an `fn iter()` that was added to `domain` branch `byo-signing-sorter-impl` (from https://github.com/NLnetLabs/domain/pull/447) and the zonemd functionality that was merged in by @mozzieongit depends on `domain` branch `zonemd-from-str` (from https://github.com/NLnetLabs/domain/pull/444)... this will be sorted out as PRs are reviewed and merged.

ximon18:
> > dnst-signzone currently allows not specifying a key to sign with. In that case it just copies the original zone over into the .signed file (with no signatures ofc).
> 
> That's a bug.

This has been resolved.

src/commands/signzone.rs
ximon18: > I updated the domain branch used by this PR to use `fmt::Write` instead of `io::Write`. The code in domain is in `src/sign/records.rs` which was in released code but never compiled as it had no `mod records.rs;` statement in `src/sign/mod.rs`. There is a `src/sign/bin/signzone.rs` orphaned example that might be affected so we should still check that. I have NOT updated this PR to use the new code in the domain branch.

This is incorrect, I was confused as it was disabled in main but that was actually done after it was last released. The latest content of this PR deletes the `src/sign/bin/signzone.rs` file, but maybe we still want an example under `examples/` where the rest of the examples are.

Regarding:

> In the short term, move those functions out of domain into dnst.

I haven't done anything more on this yet, maybe I will look at this as part of cleaning up https://github.com/NLnetLabs/domain/pull/416 for review.

Philip-NLnetLabs:
I improved key selection and fixed CDS/CDNSKEY signing.

Philip-NLnetLabs:
There is an nsec3 test that fails. No clue why.

src/commands/key2ds.rs
```diff
@@ -185,7 +185,7 @@ impl Key2ds {
             let rr = Record::new(owner, class, ttl, ds);
 
             if self.write_to_stdout {
-                writeln!(env.stdout(), "{}", rr.display_zonefile(DISPLAY_KIND));
+                writeln!(env.stdout(), "{}", rr.display_zonefile(DisplayKind::Simple));
```
ximon18, line 188: Note: Using `Simple` instead of `Tabbed` here leads to different output than the original `ldns-key2ds`.

ximon18:
To follow up: the `signzone` subcommand uses `-f -` for write to stdout while the `key2ds` subcommand uses `-n` for that. We should make these consistent in the `dnst` mode of operation.

src/commands/key2ds.rs
ximon18: See existing PR https://github.com/NLnetLabs/dnst/pull/48.

src/commands/signzone.rs
```diff
@@ -0,0 +1,4157 @@
...
+        // Extract and validate the DNSKEY RRs from the loaded zone.
+        let mut found_public_keys = vec![];
+        for rr in records.iter() {
+            if let ZoneRecordData::Dnskey(dnskey) = rr.data() {
+                // Create a public key object from the found DNSKEY RR.
+                let public_key = Record::new(rr.owner(), Class::IN, Ttl::ZERO, dnskey);
+
+                found_public_keys.push(public_key);
+            }
+        }
```
ximon18, line 673: I think this code will iterate over all records in the zone looking for DNSKEYs, yet they were already found at line 659 above, so just iterate over those instead.... unless I'm missing something.

```suggestion
        if let Some(dnskey_rrset) = &dnskey_rrset {
            for rr in dnskey_rrset.iter() {
                if let ZoneRecordData::Dnskey(dnskey) = rr.data() {
                    // Create a public key object from the found DNSKEY RR.
                    let public_key = Record::new(rr.owner(), Class::IN, Ttl::ZERO, dnskey);

                    found_public_keys.push(public_key);
                }
            }
        }
```

src/commands/signzone.rs
```diff
@@ -0,0 +1,4157 @@
...
+        let mut all_dnskeys = Vec::new();
+        let empty_records: [Record<_, _>; 0] = [];
+        for r in dnskey_rrset
+            .as_ref()
+            .map_or(empty_records.iter(), |r| r.iter())
+        {
+            all_dnskeys.push(r.clone());
+        }
+        if !self.do_not_add_keys_to_zone {
+            let dnskey_ttl = dnskey_rrset.as_ref().map_or(soa_rr.ttl(), |r| r.ttl());
```
ximon18, line 828: The [original code](https://github.com/NLnetLabs/domain/pull/416/files/fcc94d299cb0cfadb88c388aec7aabaa2da12d5b..9e166c64bbac3893eac2b4b763bea4b28b2c7f91#diff-d1299cb884bc24469f78e37c88ce3432bd5ae47f427d6eebaf34bfabe74601f1L477) in domain that used to do this had a comment explaining why it used the SOA TTL as a fallback.

I think it would be good to still document this.

Perhaps add the following text? Though ideally I'd prefer to have an RFC reference to point to if you know of one that applies here...

```suggestion
            // Use the TTL of existing DNSKEY RRs for any DNSKEY RRs that we
            // add (as all RRs in an RRSET MUST have the same TTL) else use
            // the the SOA RR TTL when there are no existing DNSKEY RRs. (as
            // that is what dnssec-signzone and ldns-signzone do).
            let dnskey_ttl = dnskey_rrset.as_ref().map_or(soa_rr.ttl(), |r| r.ttl());
```

src/commands/signzone.rs
```diff
@@ -0,0 +1,4157 @@
...
+            let key_path = env.in_cwd(key_path).into_owned();
+            // Load the private key.
+            let private_key_path = Self::mk_private_key_path(&key_path);
+            let private_key = Self::load_private_key(&env.in_cwd(&private_key_path))?;
+
+            // Note: Our behaviour differs to that of the original
+            // ldns-signzone because we are unable at the time of writing to
+            // generate a public key from a private key. As such we cannot
+            // compare the key tag of any found DNSKEY RRs to that of the
+            // public key generated from the private key. Instead we attempt
```
ximon18, line 692: This comment appears to be wrong. It says for a single "found public key" and for "each private key" yet the following loop does the opposite trying to match each found public key to the one current private key.

doc/manual/source/man/ldns-signzone.rst
```diff
@@ -22,75 +23,143 @@ Arguments
...
+      corresponding to the ``.private`` key cannot be found.
+
+      Multiple keys can be specified. Key Signing Keys are used as such when
+      they are either already present in the zone, or specified in a ``.key``
+      file, and have the Secure Entry Point flag set.
+
+      Note: Unlike the original LDNS, DNSKEY algorithms marked as ``MUST NOT``
+      or ``NOT RECOMMENDED`` in table 3.1 of RFC 8624 "Algorithm
+      Implementation Requirements and Usage Guidance for DNSSEC" are NOT
+      supported.
```
ximon18, line 46:
```suggestion
      Note: Unlike the original LDNS:
        - Public keys corresponding to ``.private`` key MUST be supplied,
          either as DNSKEY RRs in the given zone or as ``.key`` files. This
          Implementation is not able to generate missing public keys.
        - DNSKEY algorithms marked as ``MUST NOT`` or ``NOT RECOMMENDED`` in
          table 3.1 of RFC 8624 "Algorithm Implementation Requirements and
          Usage Guidance for DNSSEC" are NOT supported.
```

src/commands/signzone.rs
```diff
@@ -0,0 +1,4157 @@
...
+            if !self.no_require_keys_match_apex && public_key.owner() != soa_rr.owner() {
+                return Err(format!(
+                    "Zone apex ({}) does not match the expected apex ({})",
+                    soa_rr.owner(),
+                    public_key.owner()
+                )
+                .into());
+            }
+
+            // Attempt to crate a key pair from the loaded private and public
```
ximon18, line 732:
```suggestion
            // Attempt to create a key pair from the loaded private and public
```

src/commands/signzone.rs
```diff
@@ -0,0 +1,4157 @@
...
+            // println!(
+            //     "Loaded public key with key tag {} from '{}' for private key '{}'",
+            //     public_key.key_tag(),
+            //     public_key_path.display(),
+            //     private_key_path.display()
+            // );
+        }
+
+        // First split the key into Key Signing Keys (KSK) that sign the
+        // DNSKEY RRset and Zone Signing Keys (ZSK) that sign the zone.
```
ximon18, line 762: This comment is outdated.
```suggestion
        // First split the key into Key Signing Keys (KSK) that sign apex
        // records involved in the chain of trust, and Zone Signing Keys (ZSK)
        // that sign the rest of the zone.
```

src/commands/signzone.rs
```diff
@@ -0,0 +1,4157 @@
...
+                        .insert(hashed_name, hash_info);
+                }
+            }
+            nsec3_hashes = Some(hash_provider);
+        }
+
+        let mut signing_config: SigningConfig<_, _> = match signing_mode {
+            SigningMode::HashOnly | SigningMode::HashAndSign => {
+                // LDNS doesn't add NSECs to a zone that already has NSECs or
+                // NSEC3s. It *does* add NSEC3 if the zone has NSECs. As noted in
```
ximon18, line 1068:
```suggestion
                // NSEC3s. It *does* add NSEC3s if the zone has NSECs. As noted in
```

src/commands/signzone.rs
ximon18: @Philip-NLnetLabs: Any opinion on this?

ximon18:
> I improved key selection and fixed CDS/CDNSKEY signing.

Changes look good to me. I gave some minor feedback/suggestions.

src/commands/signzone.rs
ximon18: This discussion is outdated, the code no longer has the `is_ldns` parameter.

src/commands/signzone.rs
ximon18: This could be improved later.

src/commands/signzone.rs
Philip-NLnetLabs: I think this change is correct.

src/commands/signzone.rs
ximon18: @Philip-NLnetLabs: Do you concur that this comment is incorrect?

doc/manual/source/man/ldns-signzone.rst
Philip-NLnetLabs: I think that text is now wrong. I added support for RSASHA512 to domain for some reason.

src/commands/signzone.rs
Philip-NLnetLabs: I think this is wrong (or at least confusing) for CSKs

PR #9, opened 2024-10-17
Add keygen command.

ximon18:
WIP.


src/commands/keygen.rs
```diff
@@ -0,0 +1,214 @@
...
+            .map_err(|err| format!("private key file '{base}.private' already existed: {err}"))?;
+        let mut public_key_file = File::create_new(format!("{base}.key"))
+            .map_err(|err| format!("public key file '{base}.key' already existed: {err}"))?;
+        let mut digest_file = self
+            .make_ksk
+            .then(|| File::create_new(format!("{base}.ds")))
+            .transpose()
+            .map_err(|err| format!("digest file '{base}.ds' already existed: {err}"))?;
+        #[cfg(target_family = "unix")]
+        if self.create_symlinks {
```
mozzieongit, line 106: ldns-keygen also creates a symlink for the .ds file if the key is a ksk

src/commands/keygen.rs
```diff
@@ -0,0 +1,341 @@
...
+
+        let mut parser = lexopt::Parser::from_args(args);
+
+        while let Some(arg) = parser.next()? {
+            match arg {
+                Arg::Short('a') => {
+                    algorithm = Some(parse_os_with("algorithm (-a)", &parser.value()?, |s| {
+                        AlgorithmArg::from_str(s, true)
+                    })?);
+                }
```
tertsdiepraam, line 89: ldns:
```
$ ldns-keygen -a 8 -a 13
The -a argument can only be used once
```

src/commands/keygen.rs
```diff
@@ -0,0 +1,341 @@
...
+
+                Arg::Short('s') => {
+                    create_symlinks = true;
+                }
+
+                Arg::Short('f') => {
+                    force_symlinks = true;
+                }
+
+                // TODO: '-v' version argument?
```
tertsdiepraam, line 111: Hmm I guess we need to think about that

src/commands/keygen.rs
bal-e: Ooh, thanks!  I'll check this for other kinds of options too.

src/commands/keygen.rs
bal-e: I've copied the code from #8 here.

src/commands/keygen.rs
```diff
@@ -101,14 +103,31 @@ impl LdnsCommand for Keygen {
...
+                    #[cfg(not(target_family = "unix"))]
+                    return Err("symlinks not supported outside Unix platforms".into());
+                }
+
+                Arg::Short('v') => {
+                    let version = clap::crate_version!();
+                    // NOTE: The outer version is the latest version of 'ldns-keygen' we are
+                    // compatible with.  This needs to be updated manually.
+                    println!("DNSSEC key generator version 1.8.4 (dnst version {version})");
+                    std::process::exit(0);
```
tertsdiepraam, line 138: Would you mind waiting with this? It's incompatible with https://github.com/NLnetLabs/dnst/pull/23.

src/commands/keygen.rs
bal-e: I think I implemented that just a little before you reviewed here :)

src/commands/keygen.rs
```diff
@@ -0,0 +1,380 @@
...
+        let params = self.algorithm;
+
+        // The digest algorithm is selected based on the key algorithm.
+        let digest_alg = match params.algorithm() {
+            SecAlg::RSASHA256 => DigestAlg::SHA256,
+            SecAlg::ECDSAP256SHA256 => DigestAlg::SHA256,
+            SecAlg::ECDSAP384SHA384 => DigestAlg::SHA384,
+            SecAlg::ED25519 => DigestAlg::SHA256,
+            SecAlg::ED448 => DigestAlg::SHA256,
+            _ => unreachable!(),
```
tertsdiepraam, line 247: You might be right, but I wonder if `SecAlg` makes sense here if we always ignore some variants? Shouldn't we then take `DigestAlg`?

src/commands/keygen.rs
```diff
@@ -0,0 +1,380 @@
...
+        };
+
+        // Generate the key.
+        // TODO: Attempt repeated generation to avoid key tag collisions.
+        let (secret_key, public_key) = common::generate(params)
+            .map_err(|err| format!("an implementation error occurred: {err}").into())
+            .context("generating a cryptographic keypair")?;
```
tertsdiepraam, line 254: We still need this in the other utils. Would be cool if you could look into that at some point.

src/commands/keygen.rs
```diff
@@ -0,0 +1,380 @@
...
+        let flags = if self.make_ksk { 257 } else { 256 };
```
tertsdiepraam, line 255: Is there a less magic number-y way to do this?

src/commands/keygen.rs
```diff
@@ -0,0 +1,380 @@
...
+        let public_key = Key::new(self.name.clone(), flags, public_key);
+        let digest = self
+            .make_ksk
+            .then(|| public_key.digest(digest_alg).unwrap());
```
tertsdiepraam, line 259: When does the unwrap fail?

src/commands/keygen.rs
```diff
@@ -0,0 +1,380 @@
...
+
+        // Open the appropriate files to write the key.
+        let base = format!(
+            "K{}+{:03}+{:05}",
+            self.name.fmt_with_dot(),
+            public_key.algorithm().to_int(),
+            public_key.key_tag()
+        );
+        // TODO: Adjust for how 'Env' mocks the current directory.
```
tertsdiepraam, line 268: It's on `main` so you could do this (and make all this testable)

src/commands/keygen.rs
```diff
@@ -0,0 +1,380 @@
...
+            .map_err(|err| format!("public key file '{base}.key' already existed: {err}"))?;
+        let mut digest_file = self
+            .make_ksk
+            .then(|| File::create_new(format!("{base}.ds")))
+            .transpose()
+            .map_err(|err| format!("digest file '{base}.ds' already existed: {err}"))?;
+
+        #[cfg(target_family = "unix")]
+        if self.create_symlinks {
+            if let Ok(metadata) = std::fs::symlink_metadata(".private") {
```
tertsdiepraam, line 281: I take it there's no way to just force create a symlink?

src/commands/keygen.rs
```diff
@@ -0,0 +1,380 @@
...
+
+use crate::env::Env;
+use crate::error::{Context, Error};
+use crate::parse::parse_name;
+
+use super::{parse_os, parse_os_with, Command, LdnsCommand};
+
+#[derive(Clone, Debug, Args)]
+pub struct Keygen {
+    /// The signature algorithm to generate for
```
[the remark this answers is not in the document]
tertsdiepraam: Nvm, I see it's part of the docs on the command.

src/commands/keygen.rs
```diff
@@ -0,0 +1,380 @@
...
+            format!("error while writing private key file '{base}.private': {err}")
+        })?;
+        public_key_file
+            .sync_all()
+            .map_err(|err| format!("error while writing public key file '{base}.key': {err}"))?;
+        if let Some(digest_file) = digest_file.as_mut() {
+            digest_file
+                .sync_all()
+                .map_err(|err| format!("error while writing digest file '{base}.ds': {err}"))?;
+        }
```
tertsdiepraam, line 359: Is this really necessary?

tertsdiepraam (commented):
Well done! It's not super pretty, but it's about what you can expect for so much IO. I'd love to see this tested before we merge this.

Cargo.toml
```diff
@@ -8,11 +8,19 @@ default-run = "dnst"
...
+default = ["openssl", "ring"]
+
+# Cryptographic backends
+openssl = ["domain/openssl"]
+ring    = ["domain/ring"]
+
 [dependencies]
 clap = { version = "4.3.4", features = ["derive"] }
-domain = { version = "0.10.3", git = "https://github.com/NLnetLabs/domain.git", features = [
+domain = { git = "https://github.com/NLnetLabs/domain.git", branch = "main", features = [
```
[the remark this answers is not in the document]
bal-e: I don't know whether that will be the exact version that we release.   There's no point hardcoding it here until the release actually happens.

src/commands/keygen.rs
```diff
@@ -0,0 +1,380 @@
...
+    #[arg(short = 'k')]
+    make_ksk: bool,
+
+    /// Create symlinks '.key' and '.private' to the generated keys
+    #[arg(short = 's')]
+    #[cfg(target_family = "unix")]
+    create_symlinks: bool,
+
+    /// Overwrite existing symlinks (for use with '-s')
+    #[arg(short = 'f')]
```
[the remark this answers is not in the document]
bal-e: Maybe a `--symlink` / `-s` and `--symlink=force` / `-s force` option should be used instead?  I wasn't sure how to structure these, so I just hadn't changed them.

src/commands/keygen.rs
```diff
@@ -0,0 +1,380 @@
...
+            }
+        }
+
+        let algorithm = match algorithm {
+            Some(SecAlg::RSASHA256) => GenerateParams::RsaSha256 { bits },
+            Some(SecAlg::ECDSAP256SHA256) => GenerateParams::EcdsaP256Sha256,
+            Some(SecAlg::ECDSAP384SHA384) => GenerateParams::EcdsaP384Sha384,
+            Some(SecAlg::ED25519) => GenerateParams::Ed25519,
+            Some(SecAlg::ED448) => GenerateParams::Ed448,
+            Some(_) => unreachable!(),
```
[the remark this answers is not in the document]
bal-e: Hm, I'm not sure it's necessary.  The `-a` parser is the only place `algorithm` is written to and the list of possibly-written algorithms is clear from there.

src/commands/keygen.rs
bal-e: Hm.  We could add an option to explicitly specify a digest, but these defaults are the same that ldns uses.  I suspect that using an irregular signature/digest algorithm combination is not a good idea.

src/commands/keygen.rs
bal-e: What component are you talking about?  Error context?

src/commands/keygen.rs
bal-e: Unless we add a `DnskeyFlags` type in the `domain` crate, no.  I'll leave a TODO here about it, though.

src/commands/keygen.rs
bal-e: Right, I can make this a `.expect()`.

src/commands/keygen.rs
bal-e: Ah yeah, I need to rebase now.

src/commands/keygen.rs
bal-e: Yeah, the underlying `symlink(2)` syscall doesn't support overwriting.

src/commands/keygen.rs
bal-e: Not _particularly_, but Linux's system call interface says that an I/O error encountered by write A might actually be reported when performing write B (e.g. if write A resulted in a cache being updated, but the underlying device couldn't be written to by the time write B came around).

IMO, scripts could rely on `keygen` successfully exiting to assume that a perfectly usable keypair is now available to the caller.  I want to avoid such scripts from making dangerous changes (e.g. re-signing a zone) if the keys actually turn out to be unavailable or corrupted.

src/commands/keygen.rs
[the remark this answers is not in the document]
tertsdiepraam: Makes sense to me

src/commands/keygen.rs
tertsdiepraam: Yeah, but it's not a suggestion for this PR

src/commands/keygen.rs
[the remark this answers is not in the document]
tertsdiepraam: Even that makes a useful comment, but fine.

src/commands/keygen.rs
tertsdiepraam: Oh yeah I missed that it's not 1 to 1.

src/commands/keygen.rs
tertsdiepraam: I think that's a threat model that doesn't really apply, but alright.

src/commands/keygen.rs
```diff
@@ -0,0 +1,380 @@
...
+    )]
+    algorithm: GenerateParams,
+
+    /// Generate a key signing key instead of a zone signing key
+    #[arg(short = 'k')]
+    make_ksk: bool,
+
+    /// Create symlinks '.key' and '.private' to the generated keys
+    #[arg(short = 's')]
+    #[cfg(target_family = "unix")]
```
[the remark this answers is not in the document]
bal-e: I've moved the error handling into `execute()`, so now both `ldns` and `dnst` fail identically.

src/commands/keygen.rs
```diff
@@ -0,0 +1,605 @@
...
+        verbatim_doc_comment,
+    )]
+    algorithm: GenerateParams,
+
+    /// Generate a key signing key instead of a zone signing key
+    #[arg(short = 'k')]
+    make_ksk: bool,
+
+    /// Whether to create symlinks.
+    #[arg(short, long, value_enum, num_args = 0..=1, require_equals = true, default_missing_value = "yes", default_value = "no")]
```
tertsdiepraam, line 43: Could you make the `short = 's'` and `long = "symlink"` explicit?

tertsdiepraam: This needs a massive comment to explain the rest of the attributes you put on here as well.

src/commands/keygen.rs
```diff
@@ -0,0 +1,605 @@
...
+            match arg {
+                Arg::Short('a') => {
+                    if algorithm.is_some() {
+                        return Err("cannot specify algorithm (-a) more than once".into());
+                    }
+
+                    algorithm = parse_os_with("algorithm (-a)", &parser.value()?, |s| {
+                        Ok(match s {
+                            "list" => {
+                                // TODO: Mock stdout and process exit?
```
tertsdiepraam, line 116: This is possible after the notify branch is merged, where `parse_ldns` takes `Env` as well.

src/commands/keygen.rs
```diff
@@ -0,0 +1,605 @@
...
+                }
+
+                Arg::Short('b') => {
+                    // NOTE: '-b' can be repeated; the last instance wins.
+                    bits = parse_os("bits (-b)", &parser.value()?)?;
+                }
+
+                Arg::Short('r') => {
+                    // NOTE: '-r' can be repeated; we don't use it, so the order doesn't matter.
+                    let _ = parser.value()?;
```
tertsdiepraam, line 151: Do we want to warn or error here? Sounds like something that could lead to unexpected behaviour.

src/commands/keygen.rs
```diff
@@ -0,0 +1,605 @@
...
+            name,
+        })
+    }
+}
+
+impl From<Keygen> for Command {
+    fn from(value: Keygen) -> Self {
+        Self::Keygen(value)
+    }
+}
```
tertsdiepraam, line 216: I put all of these in `commands/mod.rs`, though maybe you're right it should be here.

src/env/mod.rs
```diff
@@ -35,7 +38,72 @@ pub trait Env {
...
+                link.display(),
+                target.display(),
+            )
+            .into()
+        })
+    }
+
+    /// Create a symlink, overwriting if it already exists.
+    #[cfg(unix)]
+    fn fs_symlink_force(&self, target: impl AsRef<Path>, link: impl AsRef<Path>) -> Result<()> {
```
tertsdiepraam, line 85: I'm not sure about adding higher-level API's than `std` here.

src/commands/keygen.rs
bal-e: Do you mean warning that we don't support `-r` at all?  Perhaps that's warranted, yeah.

src/commands/keygen.rs
tertsdiepraam: I'm mostly opposed to the inconsistency

src/commands/keygen.rs
tertsdiepraam: Explicit error saying that it's not supported yeah.

src/env/mod.rs
bal-e: For the purpose of CLI programs, `std` actually provides relatively little.  In most languages, CLI codebases tend to have their own utility modules with these kinds of specialized filesystem / OS operations, often matching up with common command-line tools (e.g. `symlink_force` is basically `ln -sf`).  I think we're going to accumulate these kinds of utility functions _somewhere_, and while `Env` isn't necessarily the right place, I think it's certainly the most convenient.

tertsdiepraam: I'm confused, you said yourself that adding a lot of functions to `Env` is probably not the right way to go? Besides, it depends on the function, I think we should keep them with the tool as long as they're not used in other tools.

This kind of muddles the idea of `Env` into both low-level environment interaction a `util` module. A function that takes `Env` as first argument is just as effective.

Don't consider this blocking though, it's alright.

bal-e: I suspect symlinks will show up in other commands as well.  I'll make a `util` module that provides these functions instead.

src/commands/keygen.rs
bal-e: I think we should move each one to its respective module.  I don't know whether I should try moving them all to their modules in this PR, or to do all of them later.

src/commands/keygen.rs
```diff
@@ -0,0 +1,629 @@
...
+                            }
+
+                            "RSASHA256" | "8" => Some(SecAlg::RSASHA256),
+                            "ECDSAP256SHA256" | "13" => Some(SecAlg::ECDSAP256SHA256),
+                            "ECDSAP384SHA384" | "14" => Some(SecAlg::ECDSAP384SHA384),
+                            "ED25519" | "15" => Some(SecAlg::ED25519),
+                            "ED448" | "16" => Some(SecAlg::ED448),
+
+                            _ => {
+                                return Err(format!("Invalid value {s:?} for algorithm (-a)"));
```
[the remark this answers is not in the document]
bal-e: Oooh, nice catch!

src/commands/mod.rs
```diff
@@ -15,7 +16,33 @@ use super::error::Error;
 
 #[derive(Clone, Debug, clap::Subcommand)]
 pub enum Command {
-    /// Prints the NSEC3 hash of a given domain name
+    /// Generate a new key pair for a given domain name
+    ///
+    /// The following files will be created:
+    ///
+    /// - K<name>+<alg>+<tag>.key: The public key file
```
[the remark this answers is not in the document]
tertsdiepraam: Another option is to put it in a string and specify `help = THE_STRING` and then it won't be a doc comment thing.

src/commands/keygen.rs
```diff
@@ -63,7 +65,7 @@ pub struct Keygen {
     symlink: SymlinkArg,
 
     /// The domain name to generate a key for
-    #[arg(value_name = "domain name", value_parser = ValueParser::new(parse_name))]
+    #[arg(value_name = "DOMAIN_NAME", value_parser = ValueParser::new(parse_name))]
```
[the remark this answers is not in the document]
tertsdiepraam: Alright, let's just merge this then and then have a PR to standardize across all utils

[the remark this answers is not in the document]
bal-e: I'm happy with lowercase names too.

PR #10, opened 2024-10-21
Implement ldns-update and dnst update

tertsdiepraam:
I guess documentation is a bit lacking and @wtoorop has updated the docs for ldns, which I should still incorporate. But the actual code is ready for review.

### DIFF
--- a/src/commands/update.rs
+++ b/src/commands/update.rs
@@ -4,13 +4,10 @@ use std::str::FromStr;
 use bytes::Bytes;
 use domain::base::iana::{Class, Opcode, Rcode};
 use domain::base::{
-    Message, MessageBuilder, Name, Question, Record, Rtype, ToName, Ttl,
-    UnknownRecordData,
+    Message, MessageBuilder, Name, Question, Record, Rtype, ToName, Ttl, UnknownRecordData,
 };
 use domain::net::client::protocol::{TcpConnect, UdpConnect};
-use domain::net::client::request::{
-    self, RequestMessage, RequestMessageMulti, SendRequest,
-};
+use domain::net::client::request::{self, RequestMessage, RequestMessageMulti, SendRequest};
 use domain::net::client::{dgram_stream, stream, tsig};
 use domain::rdata::{Aaaa, AllRecordData, Ns, Soa, A};
 use domain::resolv::{
@@ -22,6 +19,7 @@ use domain::utils::base64;
 use octseq::Array;
 use tokio::net::TcpStream;
 
+use crate::env::Env;
 use crate::error::Error;
 
 // Clap gives `Option<T>` special handling by making the argument optional.
@@ -34,15 +32,18 @@ type OptionIpAddr = Option<IpAddr>;
 #[derive(Clone, Debug, clap::Args)]
 pub struct Update {
     /// Domain name to update
-    #[arg()]
+    #[arg(value_name = "DOMAIN NAME")]
     domain: Name<Vec<u8>>,
+
     /// IP address to associate with the given domain name.
     /// Use `none` to delete the records for the domain name.
-    #[arg(value_parser = optional_ip)]
+    #[arg(value_name = "IP", value_parser = optional_ip)]
     ip: OptionIpAddr,
+
     /// Zone to update
     #[arg(long = "zone")]
     zone: Option<Name<Vec<u8>>>,
+
     /// TSIG credentials for the UPDATE packet
     #[arg(long = "tsig", value_name = "name:key[:algo]")]
     tsig: Option<TSigInfo>,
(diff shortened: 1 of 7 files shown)


src/commands/update.rs
```diff
@@ -0,0 +1,484 @@
...
+                        let key = Key::new(
+                            tsig.algorithm,
+                            &tsig.key,
+                            tsig.name.clone(),
+                            None,
+                            None,
+                        )
+                        .unwrap();
+                        let msg =
+                            RequestMessageMulti::new(msg.clone()).unwrap();
```
ximon18, line 418: What's the reason for using Multi here? I would only expect to see that if XFR responses are expected.

tertsdiepraam: It was the only thing that worked with the types...

tertsdiepraam: That'd be great! I think there's a missing impl somewhere but forgot to report that back.

ximon18: Done.

tertsdiepraam: I'm confused, there's still `RequestMessageMulti` now right?

src/commands/update.rs
```diff
@@ -0,0 +1,481 @@
...
+
+    async fn send_update_with_tsig(
+        &self,
+        msg: RequestMessage<Vec<u8>>,
+        key: Key,
+        socket: SocketAddr,
+    ) -> Result<Message<Bytes>, domain::net::client::request::Error> {
+        let tcp_conn = TcpStream::connect(socket).await.unwrap();
+        let (connection, transport) =
+            stream::Connection::<_, RequestMessageMulti<Vec<u8>>>::new(tcp_conn);
```
tertsdiepraam, line 454: Here

tertsdiepraam:
This could probably use more tests, but it seems to mostly work?

src/commands/update.rs
```diff
@@ -0,0 +1,676 @@
...
+            .await?;
+
+        let mut nsnames = response
+            .answer()?
+            .limit_to_in::<Ns<_>>()
+            .map(|ns| Ok(ns?.data().nsdname().to_name::<Vec<u8>>()))
+            .collect::<Result<Vec<_>, Error>>()?;
+
+        // The mname should be tried first according to RFC2136 4.3
+        // so we put that nsname first in the list.
```
mozzieongit, line 333:
```suggestion
        // The MNAME should be tried first according to RFC2136 4.3
        // so we put that NSNAME first in the list.
```

src/commands/update.rs
```diff
@@ -0,0 +1,676 @@
...
+        };
+
+        let mut key;
+        let mut algorithm;
+        if let Some((k, a)) = rest.split_once(':') {
+            key = k;
+            algorithm = a;
+        } else {
+            key = rest;
+            algorithm = "hmac-sha512";
```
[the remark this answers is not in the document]
tertsdiepraam: This ldns tool doesn't have a default. Which makes me wonder if this should even have a default.

src/commands/update.rs
```diff
@@ -0,0 +1,676 @@
...
+        let Some(soa) = sections.next() else {
+            return Err("no SOA found".into());
+        };
+
+        let soa_mname: Name<Vec<u8>> = soa?.data().mname().to_name();
+
+        // Step 2 - find SOA MNAME IP address, add to resolver
+        let response = resolver
+            .query(Question::new(&soa_mname, Rtype::A, Class::IN))
+            .await?;
```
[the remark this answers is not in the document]
tertsdiepraam: Good catch!

src/commands/update.rs
```diff
@@ -0,0 +1,676 @@
...
+    ip: OptionIpAddr,
+
+    /// Zone to update
+    #[arg(long = "zone")]
+    zone: Option<Name<Vec<u8>>>,
+
+    /// TSIG credentials for the UPDATE packet
+    #[arg(long = "tsig", value_name = "name:key[:algo]")]
+    tsig: Option<TSigInfo>,
+}
```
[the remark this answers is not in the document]
tertsdiepraam: @mozzieongit and I decided to defer this and make an issue out of this.

src/commands/update.rs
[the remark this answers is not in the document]
mozzieongit: Good point, maybe you're right. 
I think in the end it doesn't matter much. For some people the default will save a few characters of typing, for others it doesn't do much, other that TSIG might fail if they forget to provide the correct algorithm. Although that could already be considered annoying.

src/commands/update.rs
```diff
@@ -0,0 +1,676 @@
...
+        let resolver = env.stub_resolver_from_conf(conf).await;
+
+        let response = resolver
+            .query(Question::new(&name, Rtype::SOA, Class::IN))
+            .await?;
+
+        let mut authority = response.authority()?.limit_to::<Soa<_>>();
+        let Some(soa) = authority.next() else {
+            return Err("no SOA record found hello".into());
+        };
```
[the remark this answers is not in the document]
mozzieongit: Ah I get it now. It's to make sure we have the correct SOA record from the primary name server (and not some potentially outdated one from a secondary).

However, I believe to see an issue here (only read the code of both ldns and dnst, not tested): If `name` is the zone apex, the SOA record won't be in the authority section, but the answer section. So this is probably a bug in the original ldns-update then.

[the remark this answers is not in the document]
tertsdiepraam: Ahh that makes sense. I suppose I'll update this to look at both then?

bal-e (approved):
Nothing's leaping out at me.  It looks pretty good!

src/parse.rs
```diff
@@ -0,0 +1,66 @@
+use std::str::FromStr;
+
+use domain::{
+    tsig::{Algorithm, KeyName},
+    utils::base64,
+};
```
mozzieongit, line 6:
```suggestion
use domain::tsig::{Algorithm, KeyName};
use domain::utils::base64;
```

mozzieongit (approved):
Looks good :+1: 
Only a tiny nit

PR #13, opened 2024-10-30
Modify Clap usage to better match LDNS

ximon18:
Exit with code 1 on argument parsing error (at least that is what `ldns-nsec3-hash` does), not 2 as Clap normally does, and handle being invoked via an ldns tool name symlink.

With this PR one can do as follows:

_(assumes that we have run `cargo build --release`)_

1. Use `dnst` as usual:
```
$ target/release/dnst 
Usage: dnst <COMMAND>

Commands:
  nsec3-hash  Print the NSEC3 hash of a given domain name
  help        Show the manual pages

Options:
  -h, --help     Print help
  -V, --version  Print version

$ target/release/dnst nsec3-hash nlnetlabs.nl
e3dbcbo05tvq0u7po4emvbu79c8vpcgk.
```

2. Use `dnst` as an LDNS command, e.g.:
```
$ ln -s target/release/dnst ldns-nsec3-hash

$ ./ldns-nsec3-hash -h
Print the NSEC3 hash of a given domain name

Usage: nsec3-hash [OPTIONS] <DOMAIN_NAME>

Arguments:
  <DOMAIN_NAME>  The domain name to hash

Options:
  -a, --algorithm <NUMBER_OR_MNEMONIC>  The hashing algorithm to use [default: 1]
  -i, --iterations <NUMBER>             The number of hash iterations [default: 1] [short aliases: t]
  -s, --salt <HEX_STRING>               The salt in hex representation [default: -]
  -h, --help                            Print help

$ ./ldns-nsec3-hash nlnetlabs.nl
e3dbcbo05tvq0u7po4emvbu79c8vpcgk.
```

This also works if the binary is renamed, so if there are platforms (Windows?) that we can't use symbolic linking but instead need N copies of the binary, one for each LDNS tool to replace, when the `dnst` binary is named after a matching LDNS alias (e.g. `ldns-nsec3-hash` as defined in this PR) then the binary will behave as that subcommand alone.

To achieve this behaviour this PR uses the Clap `multicall` feature and mixes Clap builder mode in combination with the existing usage of Clap derive mode, based on [the clap mixing builder and derive docs](https://docs.rs/clap/latest/clap/_derive/index.html#mixing-builder-and-derive-apis). This is done because with derive mode one can only set `multicall=true` at compile time but in order to fallback (as described in the Clap [multicall docs](https://docs.rs/clap/latest/clap/struct.Command.html#method.multicall)) to non-multicall usage with derive mode one would have to duplicate the `Args` struct with only different derive attributes for clap (`multicall=true` and `multicall=false`). Instead we use builder mode to do the dynamic part and leave derive mode to handle the rest which is fine to determine at compile time.

I can imagine that we might later want to use more Clap builder style e.g. if for example we want to have some arguments only added to the CLI when not invoked as an LDNS tool but just as `dnst`, but for now this is at least enough to get started enabling the subcommands to be used as LDNS tool replacements.


ximon18:
One issue with this js that the top-level --version handling isn't inherited by subcommands, so one can't currently do `./ldns-nsec3-hash --version`.

tertsdiepraam:
Way too late for me to actually look into this, but note that. I was scaffolding the multicall functionality in `update` and doing it very differently. I did it differently because the ldns tools often mal very poorly to clap, so I was doing more manual argument parsing. I'll look at PR more in depth this soon.

ximon18:
Another issue with this PR is that there is no easy way to set per command alias defaults, e.g. if I invoke the `nsec3-hash` subcommand and I want the iterations to arg to differ such that:
- As `dnst nsec3-hash` the short alias should be `-i` (with `-t` also supported as a "visible alias") and a default value of 0.
- As `ldns-nsec3-hash` the short alias should be `-t` and a default value of 1.

bal-e (commented):
I think we should take some time to consider a more manual parsing approach for the `ldns-*` commands, especially given Terts' PR on `ldns-update`.  Perhaps we want to check `argv[0]` ourselves and switch to ldns parsing functions instead.

tertsdiepraam (commented):
Code looks good, but I'm not sure it's the right approach for a couple of reasons.
1. I think we shouldn't support `dnst ldns-*`
2. I think we shouldn't support `nsec3-hash` without `dnst`
3. This assumes clap parsing for the `ldns` tools, which I wasn't able to do for `ldns-update`.

ximon18:
Thanks @tertsdiepraam. Let's put this on hold (we could also just close it) and look at what you've done for update.

ximon18:
Obsoleted by #19.

PR #17, opened 2024-11-05
Add Cargo.lock to Git.

ximon18:
As `dnst` is a binary and because we do that for other binaries such as Routinator.


bal-e (approved):
Do we need to update `.gitignore` or something too?

ximon18:
`Cargo.lock` should not be added to `.gitignore` as changes to it should be added and committed to provide a known working combination of dependency versions.

I can't think of anything else impacted by adding `Cargo.lock` to the git repo.

PR #18, opened 2024-11-06
Add context tracking for errors

bal-e:
This enhances the `Error` type to store a "context stack", a way to tell where in a program the specific error occurred.  For example, when parsing a zone file, context may be used to inform the user about the line number of a parse error.  An extension trait for `Result` has also been introduced that eliminates the need for `.map_err()` to add context.


src/error.rs
```diff
@@ -1,50 +1,76 @@
...
-        Self::from(err.to_string())
+    fn from(error: io::Error) -> Self {
+        Self::new(&error.to_string())
     }
 }
 
 impl fmt::Display for Error {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
-        fmt::Display::fmt(&self.message, f)
+        f.write_str("\x1E[31mERROR:\x1E[0m ")?;
```
tertsdiepraam, line 59: The coreutils always print their own executable name (i.e. `argv[0]`), which is very nice for scripts with multiple things happening. Might be nice to do that as well.

src/error.rs
```diff
@@ -1,50 +1,76 @@
...
-            message: Cow::Owned(message),
-        }
+    /// Add context to this error.
+    pub fn context(mut self, context: &str) -> Self {
+        self.0.context.push(context.into());
+        self
     }
 }
 
-impl From<ParseError> for Error {
```
tertsdiepraam, line 27: Why this? Not specific enough? It's been pretty handy.

src/error.rs
```diff
@@ -1,50 +1,76 @@
...
+        f.write_str("\x1E[31mERROR:\x1E[0m ")?;
+        f.write_str(&self.0.primary)?;
+        self.0.context.iter().try_for_each(|c| {
+            f.write_str("\n... while ")?;
+            f.write_str(c)
+        })
+    }
+}
+
+impl fmt::Debug for Error {
```
tertsdiepraam, line 68: Shouldn't derive work for this?

tertsdiepraam (approved):

tertsdiepraam (approved):

PR #20, opened 2024-11-12
Add `ldns` binary for easier testing of ldns tools

tertsdiepraam:
Usage:
```
cargo run --bin ldns -- ldns-nsec3-hash
```
To make it easier to run these commands without making symlinks (and also being able to invoke it with `cargo run` and not `cargo build && ./ldns-nsec3-hash` and also making it possible to run it both with and without `--release`)

Side effect: the `try_ldns_compatibility` function now takes arguments so we can unit test it.

### DIFF
--- a/src/bin/ldns.rs
+++ b/src/bin/ldns.rs
@@ -1,0 +1,17 @@
+use std::process::ExitCode;
+
+use dnst::try_ldns_compatibility;
+
+fn main() -> ExitCode {
+    let mut args = std::env::args_os();
+    args.next().unwrap();
+    let args = try_ldns_compatibility(args).expect("ldns commmand is not recognized");
+
+    match args.execute() {
+        Ok(()) => ExitCode::SUCCESS,
+        Err(err) => {
+            err.pretty_print();
+            ExitCode::FAILURE
+        }
+    }
+}
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -3,6 +3,10 @@ name = "dnst"
 version = "0.1.0"
 edition = "2021"
 
+[[bin]]
+name = "ldns"
+path = "src/bin/ldns.rs"
+
 [dependencies]
 clap = { version = "4", features = ["derive"] }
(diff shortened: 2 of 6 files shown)


src/bin/ldns.rs
```diff
@@ -0,0 +1,17 @@
+use std::process::ExitCode;
```
ximon18, line 1: Maybe add a comment at the top that this binary is for testing purposes?

src/bin/ldns.rs
```diff
@@ -0,0 +1,17 @@
...
+
+use dnst::try_ldns_compatibility;
+
+fn main() -> ExitCode {
+    let mut args = std::env::args_os();
+    args.next().unwrap();
+    let args = try_ldns_compatibility(args).expect("ldns commmand is not recognized");
+
+    match args.execute() {
```
ximon18, line 10: I'm a little uncomfortable with the exit code handling here, shouldn't this be common to both `src/main.rs` and `src/bin/ldns.rs`? In fact I was expecting the whole of `src/bin/ldns.rs::main()` to be even thinner, e.g. something like"
```rust
fn main() -> ExitCode {
    let mut args = std::env::args_os();
    args.next().unwrap();
    app(args)
}
```

ximon18: Ok, I missed something @tertsdiepraam pointed out that my approach would also again let you run `dnst` commands which is a bit odd. The risk is low of this code changing often and if it becomes outdated compare to `src/main.rs` the testing would be impacted, not the actual application code so ... ignore me.

src/bin/ldns.rs
ximon18: How about:

```suggestion
//! This binary is intended to enable easy testing of ldns-xxx command
//! variants, it is not intended to be included in released versions of dnst.
use std::process::ExitCode;
```

mozzieongit (approved):
LGTM

PR #22, opened 2024-11-13
Initial packaging attempt.

ximon18:


pemensik:
Comment about licenses. Fedora has switched to SPDX licenses only. license tag should contain SPDX expression even for RPM. afaik. We have rust2rpm tool for autogenerating rpm spec file for rust projects. Those are a way to get official package into Fedora, at least. But requires each dependency properly packaged in distribution in compatible versions. Which blocks for example review of dnsi tool.

pemensik:
There is no need for *you* to prepare any packages. It is just important that package contain all required files for packagers, like me, to prepare a package. And eventually maintain it in official distribution repository.

That is not in conflict with your prepared packages.

ximon18:
If we can adjust the fields to be compatible with external needs while not breaking our existing processes we will be happy to do so.

ximon18:
There's even https://packages.fedoraproject.org/pkgs/rust-routinator/routinator/.

ximon18:
> But requires each dependency properly packaged in distribution in compatible versions. Which blocks for example review of dnsi tool.

If I understand correctly you are saying that each Rust dependency needs its own package in Fedora. I have no opinion about that approach to packaging and appreciate that distribution maintainers have their reasons for choosing this approach. 

What it is not clear to me (and my apologies, I have not read up on the Fedora Rust packaging requirements) is whether something is expected from us to get those dependencies packaged.

Also, did you mean to say dns_i which is a different project than dns_t? (the names are not very different so it could be a typo)

pemensik:
If you are interested, here are Rust guidelines: https://docs.fedoraproject.org/en-US/packaging-guidelines/Rust/
No, I am aware *dnsi* differs from *dnst*.

dnst has much lighter dependencies, in fact since we have rust-domain in Fedora already, it might become normal package if review is done.

You can check dnsi dependencies at my copr: https://copr.fedorainfracloud.org/coprs/pemensik/rust/

Yes, preparing truly shared packages is not a small task, especially when some package conflicts with other dependencies sometime.

ximon18:
So, looking further down `Cargo.toml` we do set the license to a Fedora compatible `BSD` value for use by `cargo-generate-rpm`.

Looking at https://doc.rust-lang.org/cargo/reference/manifest.html#the-license-and-license-file-fields it requires an SPDX license name, and given that we successfully publish Routinator to crates.io with the same license field value that `dnst` has in this PR, I'm confused as to what the problem is that you are asking us to fix?

Cargo.toml
```diff
@@ -19,4 +24,43 @@ lexopt    = "0.3.0"
...
+# Instead create and remove them using maintainer scripts.
+#preserve-symlinks = true
+maintainer-scripts = "pkg/debian"
+conflicts = "ldnsutils"
+
+[package.metadata.generate-rpm]
+# "BSD" alone is the 3-clause license. Inheriting "license" from above causes rpmlint to
+# complain with "invalid-license".
+# See: https://fedoraproject.org/wiki/Licensing:Main?rd=Licensing
+license = "BSD"
```
pemensik, line 53: Here, line 53 of Cargo.toml, contains BSD. That is not SPDX license. recent rpmlint should not complain about BSD-3-Clause anymore. It should complain about BSD now. Just remove extra license for rpm?

Cargo.toml
```diff
@@ -19,4 +24,43 @@ lexopt    = "0.3.0"
...
+# Perhaps a newer version of cargo-deb than the one used by Ploutos would fix this.
+# Instead create and remove them using maintainer scripts.
+#preserve-symlinks = true
+maintainer-scripts = "pkg/debian"
+conflicts = "ldnsutils"
+
+[package.metadata.generate-rpm]
+# "BSD" alone is the 3-clause license. Inheriting "license" from above causes rpmlint to
+# complain with "invalid-license".
+# See: https://fedoraproject.org/wiki/Licensing:Main?rd=Licensing
```
pemensik, line 52: That has changed and is now invalid. SPDX should be there as well. Use recent enough rpmlint. https://docs.fedoraproject.org/en-US/packaging-guidelines/LicensingGuidelines/#_license_field

Cargo.toml
pemensik: https://github.com/NLnetLabs/routinator/pull/979

tertsdiepraam:
This PR is now free of "other changes" such as fixes for nsec3-hash and fuzzing. The former were either outdated or already on main. The latter is also outdated and in a separate PR already, so duplicating it here is not necessary.

doc/manual/build/man/dnst-key2ds.1
```diff
@@ -0,0 +1,80 @@
+.\" Man page generated from reStructuredText.
```
Philip-NLnetLabs, line 1: Do we need to check in generated man pages?

AlexanderBand (approved):

src/env/mod.rs
```diff
@@ -141,13 +141,14 @@ impl<E: Env> Env for &E {
 
     fn dgram(
         &self,
-        socket: SocketAddr,
+        src: SocketAddr,
```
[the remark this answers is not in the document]
ximon18: Merging with `main` resolved it.

doc/manual/build/man/dnst-key2ds.1
ximon18: As it currently stands yes.

PR #23, opened 2024-11-14
Introduce an Env trait to mock the outside world

tertsdiepraam:
This mocks stdout, stderr and the arguments so that we can easily unit test the tools as long as they do their environment interaction via `Env`. This is the simplest implementation I could come up with, but we can provide more convenience methods and macros later.

`RealEnv` is the implementation of `Env` that provides the real stdout, stderr and `std::env::args_os`. `FakeEnv` is defined by the tests and mocks all these.

This `Env` will also include Stelline, but there's no merged utility that could use that yet, so I left it out of this PR.

Update: this PR now also includes some utility things for easier testing, such as `FakeCmd` and `FakeResult`


src/main.rs
```diff
@@ -1,18 +1,6 @@
...
-
-    match args.execute() {
-        Ok(()) => ExitCode::SUCCESS,
-        Err(err) => {
-            err.pretty_print();
-            ExitCode::FAILURE
-        }
-    }
+    let env = dnst::env::real::RealEnv;
+    ExitCode::from(dnst::run(env))
```
ximon18, line 5: Why not this?

```suggestion
    dnst::run(env).into()
```

ximon18: In fact, why not use `ExitCode` instead of `u8`?

tertsdiepraam: Question 1: sure!

Question 2: `ExitCode` implements _nothing_, not `PartialEq`, not `TryInto<u8>`, just nothing except `Termination`. So we can't use it for a function that we want to check the output of.

src/error.rs
```diff
@@ -38,8 +53,17 @@ impl Error {
...
 
     /// Pretty-print this error.
-    pub fn pretty_print(self) {
+    pub fn pretty_print(&self, env: impl Env) {
         use std::io::IsTerminal;
+        let mut err = env.stderr();
+
+        let error = match &self.0.primary {
+            PrimaryError::Clap(e) => {
+                let _ = writeln!(err, "{}", e.render().ansi());
```
ximon18, line 62: I need to try this out to see how the errors look, I find it a bit unexpected that there are two separate ways to render errors here and that they may not render the same way.

src/env/real.rs
```diff
@@ -0,0 +1,38 @@
...
+use std::fmt;
+use std::io;
+
+use super::Env;
+
+/// Use real I/O
+pub struct RealEnv;
+
+impl Env for RealEnv {
+    // fn make_connection(&self) {
```
ximon18, line 11: As you don't define the commented out `stdin()` from the trait here you could also omit these commented out functions everywhere for the moment except in the trait itself.

src/env/fake.rs
```diff
@@ -0,0 +1,145 @@
...
+
+    pub fn get_stderr(&self) -> String {
+        let r: &RefCell<_> = &self.stdout.0;
+        r.borrow().clone()
+    }
+}
+
+/// A type to used to mock stdout and stderr
+#[derive(Clone, Default)]
+pub struct FakeStream(Arc<RefCell<String>>);
```
bal-e, line 131: Hm, I feel like we can improve this a bit (e.g. keeping a `Vec` of lines so that Unix/Windows line endings don't matter, which could also improve performance).  Doesn't matter right now, though.

src/env/mod.rs
```diff
@@ -0,0 +1,48 @@
...
+pub mod fake;
+
+pub trait Env {
+    // /// Make a network connection
+    // fn make_connection(&self);
+
+    // /// Make a new [`StubResolver`]
+    // fn make_stub_resolver(&self);
+
+    fn args_os(&self) -> impl Iterator<Item = OsString>;
```
bal-e, line 16: Documentation comment :)

src/error.rs
tertsdiepraam: Clap errors are tricky because they do their own styling including their own "error" marker and colors. So we have to print it differently. The context also doesn't make sense. But I wanted to include it in the error type so that we can check for errors in tests (instead of exiting).

src/error.rs
```diff
@@ -52,9 +76,17 @@ impl Error {
...
-            eprint!("\n... while {context}");
+            let _ = writeln!(err, "\n... while {context}");
+        }
+    }
+
+    pub fn exit_code(&self) -> u8 {
+        if let PrimaryError::Clap(_) = self.0.primary {
+            2
+        } else {
+            1
```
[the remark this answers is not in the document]
tertsdiepraam: We're printing the clap error ourselves instead of returning it from main or calling `exit`. However, I'll use the `exit_code` method here.

src/env/real.rs
tertsdiepraam: Done!

bal-e (commented):
Nicely done, @tertsdiepraam!

src/error.rs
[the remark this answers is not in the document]
tertsdiepraam: Race condition!

src/env/fake.rs
```diff
@@ -0,0 +1,145 @@
...
+///
+/// This is used for testing the utilities, running the real code in a fake
+/// environment.
+#[derive(Clone)]
+pub struct FakeCmd {
+    /// The (sub)command to run, including `argv[0]`
+    pub cmd: Vec<OsString>,
+
+    /// The arguments for the commands
+    pub args: Vec<OsString>,
```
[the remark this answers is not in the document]
tertsdiepraam: The idea was that you could create a command and call it with different arguments. But since `args` now extends the arguments this can be undone I suppose.

src/env/real.rs
```diff
@@ -0,0 +1,30 @@
...
+/// Use real I/O
+pub struct RealEnv;
+
+impl Env for RealEnv {
+    fn args_os(&self) -> impl Iterator<Item = OsString> {
+        std::env::args_os()
+    }
+
+    fn stdout(&self) -> impl fmt::Write {
+        FmtWriter(io::stdout().lock())
```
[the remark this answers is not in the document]
tertsdiepraam: Didn't seem to work before, but now it does 😄

src/error.rs
ximon18: Ok.

ximon18:
As @bal-e is looking at the `FakeEnv` bit that I didn't get to yet I'm going to approve this based on the rest that I have looked at.

ximon18 (approved):

bal-e (approved):
LGTM!

PR #25, opened 2024-11-14
Updated `nsec3-hash` man pages.

ximon18:

PR #26, opened 2024-11-14
Add manual pages

mozzieongit:
Adds manual pages for the six commands based on the help output (of dnst and ldns) and the ldns manuals. I only included arguments actually supported by the dnst reimplementation. Once everything is ready, this will very likely need updating.

This also incorporates and updates the changes from #25.

### DIFF
--- a/doc/manual/source/index.rst
+++ b/doc/manual/source/index.rst
@@ -1,7 +1,12 @@
 dnst |version|
 ==============
 
-The manual goes here ...
+**dnst** is a DNS administration toolbox. It offers DNS and DNSSEC related
+functions like key generation, zone signing, printing NSEC3 hashed domain
+names, and sending UPDATE or NOTIFY messages to your name servers. More is
+coming soon.
+
+It depends on OpenSSL for its cryptography related functions.
 
 .. toctree::
    :maxdepth: 2
--- a/doc/manual/source/index.rst
+++ b/doc/manual/source/index.rst
@@ -16,6 +16,13 @@ The manual goes here ...
    man/dnst-nsec3-hash
    man/dnst-signzone
    man/dnst-update
+
+.. toctree::
+   :maxdepth: 2
+   :hidden:
+   :caption: LDNS Tools reference
+   :name: toc-reference-ldns
+
    man/ldns-key2ds
    man/ldns-keygen
    man/ldns-notify
--- a/doc/manual/source/index.rst
+++ b/doc/manual/source/index.rst
@@ -10,5 +10,15 @@ The manual goes here ...
    :name: toc-reference
    
    man/dnst
+   man/dnst-key2ds
+   man/dnst-keygen
+   man/dnst-notify
    man/dnst-nsec3-hash
-
+   man/dnst-signzone
+   man/dnst-update
+   man/ldns-key2ds
+   man/ldns-keygen
+   man/ldns-notify
+   man/ldns-nsec3-hash
+   man/ldns-signzone
+   man/ldns-update
--- a/doc/manual/source/man/dnst-key2ds.rst
+++ b/doc/manual/source/man/dnst-key2ds.rst
@@ -1,0 +1,42 @@
+dnst-key2ds
+===============
+
+Synopsis
+--------
+
+:program:`dnst key2ds` [``OPTIONS``] ``<KEYFILE>``
+
+Description
+-----------
+
+**dnst key2ds** generates a DS RR for each DNSKEY in ``<KEYFILE>``.
+
+The following file will be created for each key: ``K<name>+<alg>+<id>.ds``. The
+base name ``K<name>+<alg>+<id>`` will be printed to stdout.
+
+
+Options
+-------
+
+.. option:: -a <NUMBER OR MNEMONIC>, --algorithm <NUMBER OR MNEMONIC>
+
+      Use the given algorithm for the digest. Defaults to the digest algorithm
+      used for the DNSKEY, and if it can't be determined SHA-1.
+
+.. might change to --ignore-sep when implemented
+.. option:: -f
+
+      Ignore the SEP flag and make DS records for any key.
+
+.. option:: -n
+
+      Write the generated DS records to stdout instead of a file.
+
+.. option:: -h, --help
+
+      Print the help text (short summary with ``-h``, long help with
+      ``--help``).
+
+.. option:: -V, --version
+
+      Print the version.
--- a/doc/manual/source/man/dnst-key2ds.rst
+++ b/doc/manual/source/man/dnst-key2ds.rst
@@ -23,8 +23,11 @@ Options
       Use the given algorithm for the digest. Defaults to the digest algorithm
       used for the DNSKEY, and if it can't be determined SHA-1.
 
-.. might change to --ignore-sep when implemented
-.. option:: -f
+.. option:: -f, --force
+
+      Overwrite existing ``.ds`` files.
+
+.. option:: --ignore-sep
 
       Ignore the SEP flag and make DS records for any key.
 
--- a/doc/manual/source/man/dnst-key2ds.rst
+++ b/doc/manual/source/man/dnst-key2ds.rst
@@ -1,10 +1,10 @@
-dnst-key2ds
+dnst key2ds
 ===============
 
 Synopsis
 --------
 
-:program:`dnst key2ds` [``OPTIONS``] ``<KEYFILE>``
+:program:`dnst key2ds` ``[OPTIONS]`` ``<KEYFILE>``
 
 Description
 -----------
--- a/doc/manual/source/man/dnst-keygen.rst
+++ b/doc/manual/source/man/dnst-keygen.rst
@@ -1,10 +1,10 @@
-dnst-keygen
+dnst keygen
 ===============
 
 Synopsis
 --------
 
-:program:`dnst keygen` [``OPTIONS``] ``-a <ALGORITHM>`` ``<DOMAIN NAME>``
+:program:`dnst keygen` ``[OPTIONS]`` ``-a <ALGORITHM>`` ``<DOMAIN NAME>``
 
 Description
 -----------
--- a/doc/manual/source/man/dnst-keygen.rst
+++ b/doc/manual/source/man/dnst-keygen.rst
@@ -1,0 +1,78 @@
+dnst-keygen
+===============
+
+Synopsis
+--------
+
+:program:`dnst keygen` [``OPTIONS``] ``-a <ALGORITHM>`` ``<DOMAIN NAME>``
+
+Description
+-----------
+
+**dnst keygen** generates a new key pair for a given domain name.
+
+The following files will be created:
+
+- ``K<name>+<alg>+<tag>.key``: The public key file containing a DNSKEY RR in
+  zone file format.
+
+- ``K<name>+<alg>+<tag>.private``: The private key file containing the private
+  key data fields in BIND's *Private-key-format*.
+
+- ``K<name>+<alg>+<tag>.ds``: The public key digest file containing the DS RR
+  in zone file format. It is only created for key signing keys.
+
+| ``<name>`` is the fully-qualified owner name for the key (with a trailing dot).
+| ``<alg>`` is the algorithm number of the key, zero-padded to 3 digits.
+| ``<tag>`` is the 16-bit tag of the key, zero-padded to 5 digits.
+
+Upon completion, ``K<name>+<alg>+<tag>`` will be printed.
+
+Options
+-------
+
+.. option:: -a <MNEMONIC>
+
+      Use the given signing algorithm.
+
+      Possible values:
+
+      .. table::
+            :widths: auto
+
+            ==================== ========================
+            **Mnemonic**         **Description**
+            ==================== ========================
+            ``list``             List all algorithms
+            ``RSASHA256``        RSA with SHA-256
+            ``ECDSAP256SHA256``  ECDSA P-256 with SHA-256
+            ``ECDSAP384SHA384``  ECDSA P-384 with SHA-384
+            ``ED25519``          ED25519
+            ``ED448``            ED448
+            ==================== ========================
+
+
+.. option:: -k
+
+      Generate a key signing key (KSK) instead of a zone signing key (ZSK).
+
+.. option:: -b <BITS>
+
+      The length of the key (for RSA keys only). Defaults to 2048.
+
+.. option:: -r <DEVICE>
+
+      The randomness source to use for generation. Defaults to ``/dev/urandom``.
+
+.. option:: -s
+
+      Create symlinks ``.key`` and ``.private`` to the generated keys.
+
+.. option:: -f
+
+      Overwrite existing symlinks (for use with ``-s``).
+
+.. option:: -h, --help
+
+      Print the help text (short summary with ``-h``, long help with
+      ``--help``).
--- a/doc/manual/source/man/dnst-keygen.rst
+++ b/doc/manual/source/man/dnst-keygen.rst
@@ -35,22 +35,9 @@ Options
 
       Use the given signing algorithm.
 
-      Possible values:
-
-      .. table::
-            :widths: auto
-
-            ==================== ========================
-            **Mnemonic**         **Description**
-            ==================== ========================
-            ``list``             List all algorithms
-            ``RSASHA256``        RSA with SHA-256
-            ``ECDSAP256SHA256``  ECDSA P-256 with SHA-256
-            ``ECDSAP384SHA384``  ECDSA P-384 with SHA-384
-            ``ED25519``          ED25519
-            ``ED448``            ED448
-            ==================== ========================
-
+      Possible values are ``RSASHA256``, ``ECDSAP256SHA256``,
+      ``ECDSAP384SHA384``, ``ED25519``, ``ED448``; or ``list`` to list all
+      available algorithms.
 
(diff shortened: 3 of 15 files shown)


AlexanderBand:
Maybe "The manual goes here ..." in `index.rst` should be replaced with something describing the project. I'm not sure what but perhaps these pages can provide some inspiration:
https://www.nlnetlabs.nl/projects/ldns/about/
https://www.nlnetlabs.nl/documentation/ldns/index.html

tertsdiepraam:
I think we should merge this because this is fantastic and we can fix all my nitpicky issues later, but...

**_NITPICK MODE ACTIVATED_**

I don't like how the synopsis renders as:

> dnst [`options`] `command` [`args`]

I'd expect

> `dnst [options] <command> [args]

The main `dnst` docs should also mention how the `ldns` tools fit into this and how to use them with symlinks.

The styling is sometimes inconsistent:
- `ldns-nsec3-hash` doesn't use `<>` around values
- `ldns-keygen` is also missing some `<>` and uses lowercase
- A few places can put some more in code blocks, like `K<name>+<alg>+<id>` in `ldns-keygen`

I think that the `ldns` tools should refer to the `dnst` tools to motivate people to update their scripts.

**_NITPICK MODE DEACTIVATED_**

That being said, these are all written very well. Amazing!

tertsdiepraam:
More nitpicks:

- I don't think the arguments should be described in Options. Maybe they should have a separate heading like clap does:

```

File: doc/manual/source/man/dnst-keygen.rst [hunk 1]
```diff
@@ -0,0 +1,74 @@
...
+.. option:: -a <NUMBER OR MNEMONIC>
+
+      Use the given signing algorithm.
+
+      Possible values are:
+
+      =================== ========== =========================
+      **Mnemonic**        **Number** **Description**
+      =================== ========== =========================
+      ``list``                       List available algorithms
```
tertsdiepraam, line 43: I had no idea we had this. Cool!

doc/manual/source/man/dnst-notify.rst
```diff
@@ -0,0 +1,50 @@
...
+
+.. option:: -z <ZONE>
+
+      The zone to send the NOTIFY for.
+
+.. option:: -s <SOA VERSION>
+
+      SOA version number to include in the NOTIFY message.
+
+.. option:: -y, --tsig <name:key[:algo]>
```
tertsdiepraam, line 30: Should `name`, `key`, `algo` be uppercase?

doc/manual/source/man/ldns-nsec3-hash.rst
```diff
@@ -0,0 +1,29 @@
+ldns-nsec3-hash
+===============
+
+Synopsis
+--------
+
+:program:`ldns-nsec3-hash` :samp:`<{domain-name}>`
```
tertsdiepraam, line 7: Somewhere else you changed the `:samp:` to just uppercase with backticks.

doc/manual/source/man/ldns-signzone.rst
```diff
@@ -0,0 +1,99 @@
...
+.. option:: -f <FILE>
+
+      Write signed zone to file. Use ``-f -`` to output to stdout. Defaults to
+      ``<ZONEFILE>.signed``.
+
+.. option:: -i <DATE>
+
+      Set the inception date. Defaults to now.
+
+.. option:: -o <domain>
```
tertsdiepraam, line 55: Should this be uppercase?

doc/manual/source/man/ldns-signzone.rst
```diff
@@ -0,0 +1,99 @@
...
+-----------
+
+**ldns-signzone** signs the zone with the given key(s).
+
+Keys must be specified by their base name (usually ``K<name>+<alg>+<id>``),
+i.e. WITHOUT the ``.private`` or ``.key`` extension. Both ``.private`` and
+``.key`` files are required.
+
+A date can be a timestamp (seconds since the epoch), or of the form
+<YYYYMMdd[hhmmss]>.
```
tertsdiepraam, line 19:
```suggestion
``<YYYYMMdd[hhmmss]>``.
```

doc/manual/source/man/ldns-nsec3-hash.rst
```diff
@@ -0,0 +1,29 @@
...
+
+Description
+-----------
+
+**ldns-nsec3-hash** is used to print out the NSEC3 hash for the given domain name.
+
+Options
+-------
+
+.. option:: -a number
```
tertsdiepraam, line 17: Uppercase and with `<>`? Same for salt and count.

doc/manual/source/man/dnst-key2ds.rst
```diff
@@ -0,0 +1,42 @@
+dnst-key2ds
+===============
+
+Synopsis
+--------
+
+:program:`dnst key2ds` [``OPTIONS``] ``<KEYFILE>``
```
tertsdiepraam, line 7: Throughout the docs it's at least inconsistent whether its ```[``OPTIONS``]``` or ``` ``[OPTIONS]`` ```.

In my opinion, it should be like in the uutils docs (which makes sense that I like it because I made it):

image

Just one big code block with all possible invocations.

doc/manual/source/man/dnst-key2ds.rst
```diff
@@ -0,0 +1,42 @@
+dnst-key2ds
```
tertsdiepraam, line 1: I wonder if we should write it without `-` here?

doc/manual/source/man/ldns-keygen.rst
```diff
@@ -0,0 +1,59 @@
...
+Synopsis
+--------
+
+:program:`ldns-keygen` [``OPTIONS``] ``<DOMAIN NAME>``
+
+Description
+-----------
+
+**ldns-keygen** is used to generate a private/public keypair. When run, it will
+create 3 files; a .key file with the public DNSKEY, a .private file with the
```
tertsdiepraam, line 13: For dnst signzone, you put the `.key`, `.private` etc. in backticks.

doc/manual/source/man/dnst-signzone.rst
```diff
@@ -0,0 +1,108 @@
...
+-----------
+
+**dnst signzone** signs the zone with the given key(s).
+
+Keys must be specified by their base name (usually ``K<name>+<alg>+<id>``),
+i.e. WITHOUT the ``.private`` or ``.key`` extension. Both ``.private`` and
+``.key`` files are required.
+
+A date can be a timestamp (seconds since the epoch), or of the form
+<YYYYMMdd[hhmmss]>.
```
tertsdiepraam, line 19:
```suggestion
``<YYYYMMdd[hhmmss]>``.
```

tertsdiepraam (commented):
Something for later: we should add examples!

doc/manual/source/index.rst
```diff
@@ -10,5 +10,15 @@ The manual goes here ...
...
    man/dnst-nsec3-hash
-
+   man/dnst-signzone
+   man/dnst-update
+   man/ldns-key2ds
+   man/ldns-keygen
+   man/ldns-notify
+   man/ldns-nsec3-hash
+   man/ldns-signzone
+   man/ldns-update
```
tertsdiepraam, line 24: I think these should be in a separate `toctree`.

doc/manual/source/man/dnst-notify.rst
mozzieongit: For consistency, probably yes.

doc/manual/source/man/dnst-key2ds.rst
ximon18: Agreed, there is no `dnst-key2ds` command.

doc/manual/source/man/ldns-nsec3-hash.rst
mozzieongit: Oh, I thought I changed that. Let me change that to again to backticks ^^ 
(we're not really using the variable feature offered by `:samp:` so I changed it to be consistent with the rest)

doc/manual/source/man/dnst-signzone.rst
```diff
@@ -0,0 +1,108 @@
...
+
+      Specify the hashing algorithm. Defaults to SHA-1.
+
+.. option:: -t <NUMBER>
+
+      Set the number of hash iterations. Defaults to 0.
+
+.. option:: -s <STRING>
+
+      Specify the salt. Defaults to ``-``, meaning no salt.
```
ximon18, line 99: I think there needs to be some mention here that the string must be hex encoded.

doc/manual/source/man/dnst-key2ds.rst
mozzieongit: > Throughout the docs it's at least inconsistent whether its [``OPTIONS``] or ``[OPTIONS]``.

I'll solve the inconsistencies.

The big code block with possible invocations sound nice, but might not be able to implement that today.

doc/manual/source/man/dnst-key2ds.rst
mozzieongit: Yes, for read the docs, definitetly. For the man pages it doesn't seem to make a difference (as far as I can tell from a quick test)

doc/manual/source/man/dnst-signzone.rst
```diff
@@ -0,0 +1,108 @@
...
+
+Synopsis
+--------
+
+:program:`dnst signzone` [``OPTIONS``] ``<ZONEFILE>`` ``<KEY>...``
+
+Description
+-----------
+
+**dnst signzone** signs the zone with the given key(s).
```
ximon18, line 12: For consistency and accuracy I think this should say `zonefile` like we use below.

doc/manual/source/index.rst
mozzieongit: I made a quick change to it, but am unsure about it. When in the dnst tree, it hides the ldns compat tools. So the ldns reference is only visible when in the home/startpage. Please have a look when I push it, if that's ok behaviour. Otherwise I'll look up how to show multiple toctrees next week.

Also how should the ldns reference be called? "LDNS Tools Reference" or "LDNS Compatibility Reference" or something else?

doc/manual/source/man/dnst-signzone.rst
```diff
@@ -0,0 +1,108 @@
...
+
+      The zonefile to sign.
+
+.. option:: <KEY>...
+
+      The keys to sign the zone with.
+
+.. option:: -b
+
+      Use a more readable layout in the signed zone file and print comments on
```
ximon18, line 35: The layout of `dnst` output is NOT affected by this option, it only adds comments. The original `ldns-signzone` sets an internal flag `LDNS_COMMENT_LAYOUT` which has an effect in various places in the ldns library code on adding yet more comments (e.g. `; <void>`, `;; Empty non-terminal: <name>` and `;; Zone: <zone name>`) and adds some line breaks, but these effects have not been completely replicated in the `dnst` version yet.

ximon18: Further, per the original `ldns-signzone` man page: "Without this option, only DNSKEY RR's will have their Key Tag annotated in the comment text." - this IS implemented, but reads from the current `dnst` man page text as if no comments are ouput unless `-b` is used which is not true.

doc/manual/source/man/dnst-signzone.rst
```diff
@@ -0,0 +1,108 @@
...
+      DNSSEC records.
+
+.. option:: -d
+
+      Do not add used keys to the resulting zone file.
```
ximon18, line 40:
```suggestion
      Do not add used keys to the resulting zonefile.
```

doc/manual/source/man/dnst-signzone.rst
```diff
@@ -0,0 +1,108 @@
...
+
+.. option:: -e <DATE>
+
+      Set the expiration date. Defaults to 4 weeks from now.
```
ximon18, line 44: As is said in the original man page: "of signatures".

doc/manual/source/man/dnst-signzone.rst
tertsdiepraam: I guess technically it signs the _zone in the zonefile_?

doc/manual/source/man/dnst-signzone.rst
```diff
@@ -0,0 +1,108 @@
...
+
+.. option:: -f <FILE>
+
+      Write signed zone to file. Use ``-f -`` to output to stdout. Defaults to
+      ``<ZONEFILE>.signed``.
+
+.. option:: -i <DATE>
+
+      Set the inception date. Defaults to now.
```
ximon18, line 53: As is said in the original man page: "of signatures".

doc/manual/source/man/dnst-signzone.rst
```diff
@@ -0,0 +1,108 @@
...
+
+.. option:: -o <domain>
+
+      Set the origin for the zone (for zonefiles with relative names and no
```
ximon18, line 57: This is correct but perhaps it would be better to say that this is only needed for zones that both contain relative labels AND lack an `$ORIGIN` directive.

doc/manual/source/man/dnst-signzone.rst
```diff
@@ -0,0 +1,108 @@
...
+.. option:: -f <FILE>
+
+      Write signed zone to file. Use ``-f -`` to output to stdout. Defaults to
+      ``<ZONEFILE>.signed``.
+
+.. option:: -i <DATE>
+
+      Set the inception date of signatures to this date. Defaults to now.
+
+.. option:: -o <domain>
```
ximon18, line 55:
```suggestion
.. option:: -o <DOMAIN>
```

doc/manual/source/man/dnst-signzone.rst
```diff
@@ -0,0 +1,108 @@
...
+
+      Set SOA serial to the number of seconds since Jan 1st 1970.
+
+      If this would NOT result in the SOA serial increasing it will be
+      incremented instead.
+
+.. option:: -n
+
+      Use NSEC3 instead of NSEC. If specified, you can use extra options (see
+      :ref:`dnst-signzone-nsec3-options`).
```
ximon18, line 70:
```suggestion
      Use NSEC3 instead of NSEC. By default RFC 9276 best practice settings are used: SHA-1, no extra iterations, empty salt. To use different NSEC3 settings see :ref:`dnst-signzone-nsec3-options`.
```

doc/manual/source/man/dnst-signzone.rst
```diff
@@ -0,0 +1,108 @@
...
+
+
+.. _dnst-signzone-nsec3-options:
+
+NSEC3 options
+--------------------------------
+
+NSEC3 options for use with ``-n``.
+
+.. option:: -a <ALGORITHM>
```
ximon18, line 89:
```suggestion
.. option:: -a <ALGORITHM NUMBER OR MNEMONIC>
```

ximon18: I wonder if we shouldn't remove this argument entirely from `dnst` as the only supported algorithm is SHA-1.

mozzieongit: Yeah, might be a good idea, as it serves no real purpose currently.

doc/manual/source/man/dnst-signzone.rst
```diff
@@ -0,0 +1,108 @@
...
+
+      Specify the hashing algorithm. Defaults to SHA-1.
+
+.. option:: -t <NUMBER>
+
+      Set the number of hash iterations. Defaults to 0.
```
ximon18, line 98:
```suggestion
      Set the number of extra hash iterations. Defaults to 0.
```

doc/manual/source/man/dnst-signzone.rst
ximon18:
```suggestion
      Specify the salt. Defaults to ``-``, meaning empty salt.
```

doc/manual/source/man/dnst-signzone.rst
tertsdiepraam: This feels so wrong (the behaviour, not the description)

doc/manual/source/man/dnst-signzone.rst
```diff
@@ -0,0 +1,111 @@
...
+
+.. option:: -s <STRING>
+
+      Specify the salt as a hex string. Defaults to ``-``, meaning no salt.
```
ximon18, line 102:
```suggestion
      Specify the salt as a hex string. Defaults to ``-``, meaning empty salt.
```

doc/manual/source/index.rst
tertsdiepraam: I like the "tools" version you've got now!

doc/manual/source/man/dnst-signzone.rst
mozzieongit: Yes. It needs to be an absolute domain name. Also, signzone or rather domain's Name (::from_str?) turns any non-dot-terminated domain into an absolute/dot-terminated domain. So `dnst signzone -o origin` and `dnst signzone -o origin.` do the same.

mozzieongit:
Hm, the nsec3-hash default iterations for dnst is not yet merged into main. I'll change that to the current defaults in main, and the change would be done in #6 then, I suppose.

tertsdiepraam:
Alright I'll add it there!

tertsdiepraam (approved):

PR #27, opened 2024-11-15
Make the output of `Env::std{out, err}` a concrete type

tertsdiepraam:
Makes it less annoying to use :)

### DIFF
--- a/src/env/mod.rs
+++ b/src/env/mod.rs
@@ -23,15 +23,29 @@ pub trait Env {
     /// Get a reference to stdout
     ///
     /// Equivalent to [`std::io::stdout`]
-    fn stdout(&self) -> impl fmt::Write;
+    fn stdout(&self) -> Stream<impl fmt::Write>;
 
     /// Get a reference to stderr
     ///
     /// Equivalent to [`std::io::stderr`]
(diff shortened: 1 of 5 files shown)


src/env/mod.rs
```diff
@@ -23,17 +23,31 @@ pub trait Env {
...
+///
+/// This ensures that we don't have to `use` either [`std::fmt::Write`] or
+/// [`std::io::Write`]. Additionally, this `write_fmt` does not return a
+/// result. This means that we can use the [`write!`] and [`writeln`] macros
+/// without handling errors.
+pub struct Stream<T: fmt::Write>(T);
+
+impl<T: fmt::Write> Stream<T> {
+    pub fn write_fmt(&mut self, args: fmt::Arguments<'_>) {
+        self.0.write_fmt(args).unwrap();
```
ximon18, line 47: So why is this unwrap safe?

ximon18 (approved):

src/env/mod.rs
tertsdiepraam: Ah yeah I should document that.

It's safe in the same way that the `println` macro and `push_str` could technically fail but we decide that's so uncommon that we don't care, because if we can't write to stdout then it doesn't really matter what we do

PR #28, opened 2024-11-15
Add a bit more metadata to cargo file

pemensik:


ximon18:
Hi @pemensik,

Thankyou very much for taking the time to look at `dnst` and especially for contributing!

FYI this submission overlaps with work being done in PR #22 to create packages for `dnst`, see the [Cargo.toml](https://github.com/NLnetLabs/dnst/blob/771ce14a4a4c364fb61eb8358070461add26f656/Cargo.toml) from that PR.

Once we've finished work on that PR we can see what differences remain compared to your submission. You can of course also let us know if you think we should be doing anything differently in that PR.

Thanks,

Ximon

pemensik:
@ximon18 I am attempting to create rpm package using rust2rpm tool. I am not targetting any specific cargo tool. It just complained about license tag missing, so I imported basic stuff into it.

pemensik:
I made my change more atomic. Yours is too large for a single PR, IMO :)

ximon18:
> @ximon18 I am attempting to create rpm package using rust2rpm tool. I am not targetting any specific cargo tool. It just complained about license tag missing, so I imported basic stuff into it.

If I have time I will loom at whether it is possible to change the license specified in `Cargo.toml` to work with your `rust2rpm` tool as well as our Ploutos tool (which in turn uses cargo-generate-rpm).

ximon18:
> I made my change more atomic. Yours is too large for a single PR, IMO :)

I suppose that's a matter of perspective and policy. The changes in my PR are all part of a single task so should be together.

tertsdiepraam:
While #22 does overlap, this PR has a few more fields and #22 might still take a bit to merge. In the meantime, we can merge this I think.

tertsdiepraam (approved):

PR #35, opened 2024-11-25
Optionally show progress while signing.

ximon18:
Signing can be a slow operation.

Prior to this PR there is no feedback at all when signing, one doesn't even know if the signing is stuck or going so slowly that you'd rather stop it than wait. 

This PR uses support in `domain` to know how far operations have progressed in combination with the `indicatif` crate to output helpful progress bars.

Note: This is quick proof-of-concept, it is quite messy at the moment, hence the draft status.

Example appearance:
image

image


src/commands/signzone.rs
```diff
@@ -474,20 +492,48 @@ impl SignZone {
         let (apex, ttl) = Self::find_apex(&records).unwrap();
 
         // Hash the zone with NSEC or NSEC3.
+        let pb = if matches!(&log, Some(log) if log.is_terminal() && self.progress) {
+            let pb = ProgressBar::new(0).with_message("Hashing");
```
tertsdiepraam, line 496: This should probably be `ProgressBar::no_length()` and then with slightly different parsing. Also we should call `tick()` to make this actually show up, because it doesn't seem to do that by itself.

src/commands/signzone.rs
```diff
@@ -499,20 +545,97 @@ impl SignZone {
...
+            let pb = ProgressBar::new(num_families as u64);
+            pb.set_draw_target(ProgressDrawTarget::stderr_with_hz(1));
+            pb.set_style(
+                ProgressStyle::with_template(PROGRESS_STYLE)
+                    .unwrap()
+                    .with_key("eta", |state: &ProgressState, w: &mut dyn Write| {
+                        write!(w, "{:.1}s", state.eta().as_secs_f64()).unwrap()
+                    })
+                    .progress_chars("#>-"),
+            );
```
tertsdiepraam, line 633: This style should probably be saved somewhere so we can reuse it.

src/commands/signzone.rs
```diff
@@ -692,19 +868,41 @@ impl SignZone {
...
 
+        if let Some(pb) = pb {
+            let len = pb.length().unwrap();
+            let elapsed = pb.elapsed();
+            pb.finish_and_clear();
+            if self.verbose {
+                if let Some(log) = log {
+                    writeln!(
+                        log,
+                        "Loaded {len} records from {} [{} bytes] in {}",
```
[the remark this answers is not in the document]
tertsdiepraam: But `records.len()` or something similar should work.

tertsdiepraam (commented):
This is very cool, but I have different taste in a few aspects apart from the small issues I noted below.

So these are just my personal preferences stated as fact:
- This should be a `MultiProgress` with multiple steps. And we should show at which step it is.
- The progress bars should not be cleared.
- Progress bars should not be reused with different messages.
- Spinners should be used for steps where we don't know how long it takes.

src/commands/signzone.rs
```diff
@@ -474,20 +492,48 @@ impl SignZone {
...
+                        if let Some(pb) = &pb {
+                            if inc_len > 0 {
+                                pb.inc_length(inc_len as u64);
+                            }
+                            if inc_pos > 0 {
+                                pb.inc(inc_pos as u64);
+                            }
+                            if let Some(new_phase) = new_phase {
+                                pb.set_message(new_phase);
+                            }
```
tertsdiepraam, line 534: This also feels like it should create new progress bars or something.

ximon18:
> This is very cool, but I have different taste in a few aspects apart from the small issues I noted below.
> 
> 
> 
> So these are just my personal preferences stated as fact:
> 
> - This should be a `MultiProgress` with multiple steps. And we should show at which step it is.
> 
> - The progress bars should not be cleared.
> 
> - Progress bars should not be reused with different messages.
> 
> - Spinners should be used for steps where we don't know how long it takes.

Thx:
- `MultiProgress` seemed to me (a) to be about concurrent actions which these are not, and (b) likely has some overhead to be thread safe, so I chose not to use it for those reasons.
> - Keeping the progress bar.. I'm not sure. I find the summary afterward more useful and leaving a wide bar across your screen that no longer updates to tell you anything new a bit annoying.
- The different messages are phases of a single task with progress.
- I wondered maybe "Sorting" isn't appearing because the sorting operation is so CPU intensive, no I/O or anything, but didn't look into it yet. I wasn't aware of a pure spinner, I wondered about a left-right-left-right bouncing progress bar, or .. don't know, ideally sorting would also report progress as it can be slow on a big zone.

Btw, what does it mean to state preference as fact? 🤪

ximon18:
Btw, the large repetitive progress bar construction should be factored out, the code is v messy at present, but I didn't want to spend time on that yet.

src/commands/signzone.rs
ximon18: I tried calling `tick()` but that also didn't help.

tertsdiepraam:
> Btw, what does it mean to state preference as fact? 🤪

I just meant that I phrased without prefixing everything with "I think" :)

src/commands/signzone.rs
ximon18: Yup, there's far too much repetitive boilerplate here, but not being familiar with the progress bar crate I wanted to first get something working.

tertsdiepraam:
> Keeping the progress bar.. I'm not sure. I find the summary afterward more useful and leaving a wide bar across your screen that no longer updates to tell you anything new a bit annoying.

Makes sense yeah. It's kind of satisfying to me, but I understand.

src/commands/signzone.rs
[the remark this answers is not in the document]
ximon18: I can't see the surrounding code on my phone via the GH app.. I _think_ there was one place where I thought the same but was wrong, maybe you are right, but I can't check right now.

ximon18:
> > Btw, what does it mean to state preference as fact? 🤪
> 
> I just meant that I phrased without prefixing everything with "I think" :)

I'm glad I asked. I took it to mean that while it is your preference you believe it not to just be a preference but to be a fact that it should be done that way ;-)

PR #36, opened 2024-11-27
Error reporting extensions.

ximon18:
This builds on the work already done by @bal-e to:

- Determine the terminal status from the actual stream an error is written to.
- Factor out colourisation to the stream which knows if it is a terminal or not.
- Add support for yellow warnings as well as red errors.
- Add a line break at the end of written errors/warnings.

### DIFF
--- a/src/env/fake.rs
+++ b/src/env/fake.rs
@@ -51,11 +51,11 @@ impl Env for FakeEnv {
     }
 
     fn stdout(&self) -> Stream<impl fmt::Write> {
-        Stream(self.stdout.clone())
+        Stream::new(self.stdout.clone(), false)
     }
 
     fn stderr(&self) -> Stream<impl fmt::Write> {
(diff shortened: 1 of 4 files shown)


mozzieongit (commented):
(See my train of thought in the following paragraphs ^^)

I feel like this is missing the warning interactions (like writing them to stderr).

Also this seems like we're extending the error type to be basically a logging type (error vs warning and next maybe info and notice?) with context.

Would a logging module be better suited for this? In the current state, the warning seems to be stored in the error chain, but not printed, unless the commands return Error, which it hopefully doesn't, when there are no errors.

ximon18:
@tertsdiepraam: Is this obsoleted by https://github.com/NLnetLabs/dnst/pull/62?

tertsdiepraam:
That was the idea yes, if you agree.

ximon18:
Obsoleted by #62.

PR #39, opened 2024-11-28
All: add version printing mechanism via `Report` command

tertsdiepraam:
The command cannot be triggered via clap but can be returned by the routines that do ldns parsing. So they can print stuff without being executed. This allows them to print things and exit in a way that we can mock.

### DIFF
--- a/src/commands/mod.rs
+++ b/src/commands/mod.rs
@@ -6,6 +6,8 @@ pub mod nsec3hash;
 use std::ffi::{OsStr, OsString};
 use std::str::FromStr;
 
+use clap::crate_version;
+
 use crate::env::Env;
 use crate::Args;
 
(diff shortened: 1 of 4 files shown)


src/commands/mod.rs
```diff
@@ -58,6 +73,16 @@ pub trait LdnsCommand {
             Err(e) => Err(format!("Error: {e}\n\n{}", Self::HELP).into()),
         }
     }
+
+    fn report_version() -> Args {
```
bal-e, line 77: I like that individual commands can override this if they need.

src/commands/mod.rs
```diff
@@ -35,6 +44,10 @@ impl Command {
             Self::Nsec3Hash(nsec3hash) => nsec3hash.execute(env),
             Self::Key2ds(key2ds) => key2ds.execute(env),
             Self::Help(help) => help.execute(),
+            Self::Report(s) => {
+                write!(env.stdout(), "{s}");
```
bal-e, line 48: Do we want to use `writeln!()` here?  Especially for one-line reports, nobody remembers to put a newline after.

PR #48, opened 2024-12-05
Format `keygen` and `key2ds` subcommand output using tabs like `ldns-keygen` and `ldns-key2ds`.

ximon18:
This PR updates the `keygen` and `key2ds` subcommands to format their output using tabs instead of whitespace in order to match the behaviour of `ldns-keygen` and `ldns-key2ds`.


tertsdiepraam (commented):
I assume you're waiting for https://github.com/NLnetLabs/domain/pull/463, but once that's merged this looks fine too.

tertsdiepraam (approved):

mozzieongit (approved):

mozzieongit:
> Hmm, I'm wondering now why `dnst` is apparently producing a complete `DS` RR by outputting it in pieces and thus `display`ing RDATA only, rather than outputting a complete RR which I assume would not have hit this "bug" in `domain` in the first place...

Maybe because we switched from DisplayKind::Simple to DisplayKind::Tabbed?

ximon18:
I updated the title and description to better match what is actually changing.

Before merging this (as it is approved) I wonder, have we just approved the technical changes or are we also okay with this change applying in both `dnst` mode as well as`ldns` emulation mode, or should it only apply to `ldns` emulation mode?

PR #57, opened 2025-02-28
Don't send EDNS(0) OPT record with NOTIFY.

ximon18:
See https://github.com/opendnssec/opendnssec/pull/865.

Lacks a unit test.

### DIFF
--- a/src/commands/notify.rs
+++ b/src/commands/notify.rs
@@ -267,6 +267,10 @@ impl Notify {
         let mut config = dgram::Config::new();
         config.set_max_retries(self.retries);
 
+        // Prevent sending of an EDNS(0) OPT record as it causes OpenDNSSEC 2.1.14 to crash.
+        // See: https://github.com/opendnssec/opendnssec/pull/865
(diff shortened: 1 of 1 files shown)


src/commands/notify.rs
```diff
@@ -267,6 +267,10 @@ impl Notify {
         let mut config = dgram::Config::new();
         config.set_max_retries(self.retries);
 
+        // Prevent sending of an EDNS(0) OPT record as it causes OpenDNSSEC 2.1.14 to crash.
```
Philip-NLnetLabs, line 270: I'm not sure that we should explicitly mention a workaround for our own software, especially that it can cause a crash. Something generic like 'some software cannot handle an OPT record in a NOTIFY' might be better.

ximon18: As we develop in the open I don't see value in being vague.

Also if using your proposed text the reader may think "I've never heard of any software that has this issue" and proceed to remove the line of code perceiving it to be useless.

tertsdiepraam (approved):

PR #55, opened 2025-02-07
Track changes in domain related to the signing PR.

Philip-NLnetLabs:

PR #58, opened 2025-04-14
Sync with domain dnssec restructure

tertsdiepraam:
Should be mostly straightforward, but I might have made some mistakes. I'd like this to be review by @Philip-NLnetLabs at least.

This is somewhat of a blocker to https://github.com/NLnetLabs/dnst/pull/8, which should also get the same treatment. Also blocking https://github.com/NLnetLabs/dnst/pull/24 which breaks currently as well.


Philip-NLnetLabs:
I think I already got all of that the in the adaption of the signzone PR.

tertsdiepraam:
Maybe, but currently everything fails to build on main, so this is just a PR to unblock merging other stuff.

tertsdiepraam:
Nevermind, you're talking about another PR. I'll take a look there.

tertsdiepraam:
#55 actually did not merge to main, so reopening this.

bal-e (approved):
Looks good, @tertsdiepraam.

PR #59, opened 2025-04-16
Remove license-file field from Cargo.toml

tertsdiepraam:
This field should only be used for non-standard license files.

This is specified by Cargo: https://doc.rust-lang.org/cargo/reference/manifest.html#the-license-and-license-file-fields

And also by, for example, Fedora:

> For this reason, the license metadata for all Rust crates packaged for Fedora MUST match the license tag of the Fedora package itself. Any crates that set package.license-file in their metadata (which is reserved for non-standard / non-SPDX licenses) MUST be patched to set package.license in their metadata instead in cases where this is not appropriate and an accurate SPDX expression can be provided. Patches like this SHOULD be submitted upstream.

https://docs.fedoraproject.org/en-US/packaging-guidelines/Rust/#_crate_license

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -9,7 +9,6 @@ authors = ["NLnet Labs <dns-team@nlnetlabs.nl>"]
 description = "Rust reimplementation of important ldns programs."
 categories = ["command-line-utilities"]
 license = "BSD-3-Clause"
-license-file = "LICENSE"
 keywords = ["DNS", "domain", "ldns"]
 rust-version = "1.79"
 


ximon18:
I wonder if this was caused by cargo-deb or cargo-generate-rpm not working at that point unless license-file was specified... I don't recall.

tertsdiepraam:
Note that it's only there because of the other PR with cargo metadata, not because of your PR. That one didn't include a `license-file`, see this diff from before I started committing to it: https://github.com/NLnetLabs/dnst/pull/22/files/0271a6fd522b1673aad67f5c67ce3d59713a0148#diff-2e9d962a08321605940b5a657135052fbcef87b5e360662bb527c96d9a615542

Philip-NLnetLabs (approved):

tertsdiepraam:
Krill and Routinator both have the additional license data in the `deb` and `rpm` sections, so this is fine.

PR #60, opened 2025-04-16
Fix up newlines in error printing

tertsdiepraam:
The errors currently are printed without a newline, which is annoying because it requires us to add `\n` to every message. Additionally, there seemed to be a blank line between each context line.

@bal-e do you remember why you wrote it this way?

cc @Philip-NLnetLabs

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -83,9 +83,9 @@ impl Error {
             "ERROR:"
         };
 
-        write!(err, "[{prog}] {error_marker} {error}");
+        writeln!(err, "[{prog}] {error_marker} {error}");
         for context in &self.0.context {
-            writeln!(err, "\n... while {context}");
+            writeln!(err, "... while {context}");
         }
     }
 


ximon18:
By "fix" do you mean you think it is incorrect for log messages to span multiple lines (if so I agree as that can lead to hard to read logs when multiple threads log at the same time), but I _assume_ that @bal-e wrote it this way as it was more aesthetically pleasing to see the context on a separate line to the error itself?

bal-e:
I have a feeling this code was originally intended for `fmt::Display`; `Display` impls don't add a trailing newline by convention, so I was trying to match that (the original code should have been using `write!` instead of `writeln!` here anyway).  In the context of printing directly to stderr, however, your approach makes more sense.

bal-e:
The `writeln!("\n...")` is definitely a bug, because it would be leaving blank lines between every bit of context.

tertsdiepraam:
@ximon18 I updated the description.

PR #61, opened 2025-04-17
CLI Key Management

Philip-NLnetLabs:
Demonstrate key management using a CLI tool. Initial interface between key management and signer.


tertsdiepraam:
This seems to be missing the file for the `keyset` command.

mozzieongit (commented):
I know this is still a draft, but here my notes from reading this PR to understand the other PRs:

- Add comments on the subcommands, that say what they do and how to use them
  - And a note about what to do after each command? e.g. after init / creating key, mention which command should be called next and when
- Split up big run function into sub functions (obsolete by #108)
- Consistently use `env.in_cwd()` vs raw paths (e.g. `read_to_string(priv_url.path())`) or make clear why not using `env.in_cwd()`
- Add tests? I'm sure not all can be tested easily, but maybe some can
- Explain to users which TTL value is expected in the propagation/cache subcommands
- adjust the actions output after each command, to something human-actionable, so the user knows what to do with that information

src/commands/mod.rs
```diff
@@ -94,10 +99,12 @@ impl Command {
         match self {
             Self::Key2ds(key2ds) => key2ds.execute(env),
             Self::Keygen(keygen) => keygen.execute(env),
+            Self::Keyset(keyset) => keyset.execute(env),
             Self::Nsec3Hash(nsec3hash) => nsec3hash.execute(env),
             Self::Notify(notify) => notify.execute(env),
             Self::SignZone(signzone) => signzone.execute(env),
             Self::Update(update) => update.execute(env),
+            // Self::Help(help) => help.execute(env),
```
tertsdiepraam, line 107: Nit but I think this stems from a conflict.

Philip-NLnetLabs: As far as I know, help was disabled by Jannik until the help command can render man pages.

Philip-NLnetLabs: Ah, I see, in commit 6915b816a65dbafa1ed27f0cd84781362dbb7f0d:
-            // Self::Help(help) => help.execute(env),
+            Self::LdnsUpdate(ldnsupdate) => ldnsupdate.execute(env),

That was just too confusing.

PR #62, opened 2025-04-22
Use `tracing_subscriber` to print warnings and errors

tertsdiepraam:
Re-imagining of https://github.com/NLnetLabs/dnst/pull/36.

Uses `tracing` so we can catch `domain` logs as well as `dnst`'s. We can't use the `log` crate for this because that cannot split the logs based on the thread, which we need for tests.

### DIFF
--- a/src/commands/notify.rs
+++ b/src/commands/notify.rs
@@ -582,15 +582,13 @@ mod tests {
 
     #[test]
     fn invalid_domain_name() {
-        let rpl = format!(
-            "
+        let rpl = "
             CONFIG_END
 
             SCENARIO_BEGIN
 
             SCENARIO_END
-        "
-        );
+        ";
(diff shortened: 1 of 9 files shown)


tertsdiepraam:
I wouldn't call this "ready" but I'm marking it ready, because I'd like some review and feedback about whether this is the direction we want to take.

src/commands/notify.rs
```diff
@@ -587,4 +576,21 @@ mod tests {
         assert_eq!(res.exit_code, 0);
         assert!(res.stderr.contains("Name or service not known"));
     }
+
+    #[test]
```
ximon18, line 580: Not sure why this additional test is part of this PR?

ximon18 (approved):

PR #64, opened 2025-04-22
`nsec3-hash`: set default number of iterations to 0

tertsdiepraam:
This changes the default number of iterations of `ldns-nsec3-hash` ~to a more sensible default~ in accordance with latest best practice per RFC 9276. This explicitly breaks compatibility with the `ldns` version of this command.


bal-e (approved):

ximon18 (approved):
Not just more sensible but also in accordance with latest best practice per RFC 9276. This was actually earlier in the sign zone branch code but accidentally removed.

ximon18:
Fixes https://github.com/NLnetLabs/dnst/pull/8#issuecomment-2815153458

PR #67, opened 2025-05-09
Fix signing of zones with early glue.

ximon18:
- Use updated domain which now requires that the apex of the zone being signed be passed to sign_zone(), rather than it failing to determine the right zone apex itself.
- Make `find_apex()` require the found apex to be the expected one when `-o` is in use and to only match SOA RRs at the root of a zone.
- Add some tests for the case of early glue showing that the wrong apex is no longer searched for a SOA by the signer.
- Remove the no longer necessary separation of execute() into an extra go_further() fn, which was previously needed as a workaround for using the right generic values.
- Some cleanup in affected/related code, e.g. bump the SOA SERIAL and use only that bumped SOA RR rather than use two different SOA RRs, and refer to `new_default_rr_ttl` instead of `soa_rr.ttl()` to make it clear that it's not per se the SOA RR TTL we are interested in, this is just the default we choose to use.

Note: currently depends on domain PR https://github.com/NLnetLabs/domain/pull/521.

### DIFF
--- a/src/commands/signzone.rs
+++ b/src/commands/signzone.rs
@@ -27,7 +27,6 @@ use core::fmt::Write;
 use core::ops::Add;
 use core::str::FromStr;
 
-use std::cmp::min;
 use std::collections::{HashMap, HashSet};
 use std::ffi::OsString;
 use std::fmt::{self, Display};
@@ -640,21 +639,40 @@ impl SignZone {
         // Read the zone file.
         let mut records = self.load_zone(&env.in_cwd(&self.zonefile_path))?;
 
-        // Extract the SOA RR from the loaded zone.
-        let Some(soa_rr) = records.find_soa() else {
-            return Err(format!(
-                "Zone file '{}' does not contain a SOA record",
-                self.zonefile_path.display()
-            )
-            .into());
+        // Find apex records that require special processing.
+        let soa_rr = Self::find_apex(&records, self.origin.as_ref())?.clone();
+
+        // Process the SOA RR.
+        let soa_rdata = if self.set_soa_serial_to_epoch_time {
+            let new_soa_rdata = Self::mk_bumped_soa_rdata(&soa_rr);
+            records.update_data(
+                |rr| rr == &soa_rr,
+                ZoneRecordData::Soa(new_soa_rdata.clone()),
+            );
+            new_soa_rdata
+        } else {
+            // SAFETY: Already checked before this point.
+            let ZoneRecordData::Soa(soa_rdata) = soa_rr.data() else {
+                unreachable!()
+            };
+            soa_rdata.clone()
         };
-        let ZoneRecordData::Soa(_) = soa_rr.first().data() else {
-            return Err(format!(
-                "Zone file '{}' contains an invalid SOA record",
-                self.zonefile_path.display()
-            )
-            .into());
-        };
+        let soa_serial = soa_rdata.serial();
+        let apex = soa_rr.owner();
+        let zone_class = soa_rr.class();
+
+        // Use the SOA RR TTL as the TTL for any new RRs that we add for which
+        // there are otherwise no rules about what TTL to use for the RTYPE
+        // being added.
+        //
+        // Rationale:
+        // While in RFC 1033 section "RESOURCE RECORDS" it says to use the SOA
+        // MINIMUM time when the TTL to use for a new RR is unknown, neither
+        // dnssec-signzone nor ldns-signzone do that, instead they use the TTL
+        // of the SOA RR as the default, plus RFC 1033 predates RFC 1034 and
+        // it's thus unclear if it is relevant. So we will do the same as
+        // dnssec-signzone and ldns-signzone.
+        let new_rr_default_ttl = soa_rr.ttl();
 
         let dnskey_rrset = records.find_apex_rtype(soa_rr.owner(), Rtype::DNSKEY);
 
@@ -826,6 +844,9 @@ impl SignZone {
         }
         if !self.do_not_add_keys_to_zone {
             let dnskey_ttl = dnskey_rrset.as_ref().map_or(soa_rr.ttl(), |r| r.ttl());
+                .as_ref()
+                .map_or(new_rr_default_ttl, |r| r.ttl());
+
             // Make sure that the DNSKEY RRset contains all keys.
             for k in &signing_keys {
                 let pubkey = k.dnskey();
@@ -894,22 +915,6 @@ impl SignZone {
             records.insert(r).expect("should not fail");
         }
 
-        self.go_further(env, records, signing_mode, &zone_signing_keys, out_file)
-    }
-
-    #[allow(clippy::too_many_arguments)]
-    fn go_further(
-        &self,
-        env: impl Env,
-        mut records: SortedRecords<
-            StoredName,
-            ZoneRecordData<Bytes, StoredName>,
-            MultiThreadedSorter,
-        >,
-        signing_mode: SigningMode,
-        signing_keys: &[&SigningKey<Bytes, KeyPair>],
-        out_file: PathBuf,
-    ) -> Result<(), Error> {
         let mut writer = if out_file.as_os_str() == "-" {
             FileOrStdout::Stdout(env.stdout())
         } else {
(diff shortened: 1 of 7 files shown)


src/commands/signzone.rs
```diff
@@ -735,10 +753,9 @@ impl SignZone {
 
                 // Verify that the owner of the public key matches the apex of the
                 // zone.
-                if public_key.owner() != soa_rr.owner() {
+                if public_key.owner() != apex {
                     return Err(format!(
-                        "Zone apex ({}) does not match the expected apex ({})",
-                        soa_rr.owner(),
+                        "Zone apex ({apex}) does not match the expected apex ({})",
```
Philip-NLnetLabs, line 758: Should this say something like 'public key owner name does not match zone apex'?

ximon18: Yes. I'll fix it.

src/commands/signzone.rs
```diff
@@ -1418,41 +1419,73 @@ impl SignZone {
...
     }
 
-    fn find_apex(
-        records: &SortedRecords<StoredName, StoredRecordData, MultiThreadedSorter>,
-    ) -> Result<(StoredName, Class, Ttl, Serial), Error> {
-        let soa = match records.find_soa() {
-            Some(soa) => soa,
-            None => {
-                return Err(Error::from("Invalid zone file: Cannot find SOA record"));
+    // TODO: Add tests.
```
Philip-NLnetLabs, line 1422: Should this TODO stay?

src/commands/signzone.rs
```diff
@@ -2710,6 +2739,8 @@ mod test {
 
         let zone_file_path = mk_test_data_abs_path_string("test-data/example.rfc8976-simple");
 
+        // TODO: Remove outdated references to -b which the tests are no longer using.
```
Philip-NLnetLabs, line 2742: What about this TODO?

src/commands/signzone.rs
```diff
@@ -1418,41 +1419,73 @@ impl SignZone {
...
+                    }
+                }
+                return Ok(rr);
             }
-            _ => unreachable!(),
-        };
+        }
 
-        Ok((soa.owner().clone(), soa.class(), ttl, serial))
+        if let Some(expected_origin) = origin {
```
Philip-NLnetLabs, line 1474: Is this reachable?

src/commands/signzone.rs
ximon18: This isn't really anything to do with this PR but it is correct that there are comments talking about `-b` from earlier versions of the tests that used `-b` and now don't, so the comments are outdated and should be edited.

src/commands/signzone.rs
ximon18: Let me see what code coverage says.

src/commands/signzone.rs
```diff
@@ -1418,41 +1419,67 @@ impl SignZone {
...
+        // SAFETY: We just verified that records is not empty.
+        let mut iter = records.iter().peekable();
+        let mut zone_name = iter.peek().unwrap().owner();
+        while let Some(rr) = iter.next() {
+            // Only match SOA RRs at the apex of a zone. If the current name
+            // is a child of the last seen name skip it because as the records
+            // are in sorted order children appear in the iteration order
+            // after their parent.
+            if rr.owner() != zone_name {
+                if rr.owner().starts_with(zone_name) {
```
Philip-NLnetLabs, line 1448: I guess this should be ends_with?

src/commands/signzone.rs
ximon18: image

Quite a bit of code not covered by existing tests.

src/commands/signzone.rs
ximon18: That somewhat proves the point about testing as the coverage report showed this condition was not tested.

src/commands/signzone.rs
ximon18: Yes, I just didn't have time at the time.

src/commands/signzone.rs
ximon18: I'll close these testing gaps.

src/commands/signzone.rs
```diff
@@ -1418,41 +1419,67 @@ impl SignZone {
...
 
-    fn find_apex(
-        records: &SortedRecords<StoredName, StoredRecordData, MultiThreadedSorter>,
-    ) -> Result<(StoredName, Class, Ttl, Serial), Error> {
-        let soa = match records.find_soa() {
-            Some(soa) => soa,
-            None => {
-                return Err(Error::from("Invalid zone file: Cannot find SOA record"));
+    // TODO: Add tests.
+    fn find_apex<'a>(
```
ximon18, line 1423: @Philip-NLnetLabs: Hang on. We discussed yesterday that we probably don't even want this functionality as discovering the apex is prone to problems, the apex should be the origin found during zone parsing shouldn't it?

Philip-NLnetLabs: Zone parsing doesn't find the origin if the zone has no $ORIGIN and only absolute names. I don't think $ORIGIN has to be the name of the zone. You could have $ORIGIN com. and then example SOA xxxx. So the zone name would be example.com.

ximon18: Good point. It's confusing that both `ldns-signzone` and `dnssec-signzone` use `-o <origin>` to refer to what is actually I believe the zone name, not the `$ORIGIN` in a zone file.

Philip-NLnetLabs: The confusing part is that -o is optional. It should be required, then the problem is solved.

src/commands/signzone.rs
```diff
@@ -1418,41 +1419,67 @@ impl SignZone {
...
+        // SAFETY: We just verified that records is not empty.
+        let mut iter = records.iter().peekable();
+        let mut zone_name = iter.peek().unwrap().owner();
+        while let Some(rr) = iter.next() {
+            // Only match SOA RRs at the apex of a zone. If the current name
+            // is a child of the last seen name skip it because as the records
+            // are in sorted order children appear in the iteration order
+            // after their parent.
+            if rr.owner() != zone_name {
+                if rr.owner().ends_with(zone_name) {
```
Philip-NLnetLabs, line 1448: In wonder about:
com. IN A 1.2.3.4
example.com. SOA ...
example.com. NS com.

I think this would fail. Maybe avoid checking for children and just take the first SOA?

ximon18: > The confusing part is that -o is optional. It should be required, then the problem is solved.

Didn't we say yesterday that it should not be optional?

Philip-NLnetLabs: I think that is fine. We can see if people complain. Then the code needs to be changed to reflect that it is not optional.

ximon18: I have now made the zone owner name a mandatory positional parameter.

I don't like it, it feels like a regression compared to the original LDNS, however it is in line with `dnssec-signzone` which complains (even with `$ORIGIN` in the zone file at the top and the first record having an absolute owner name and being a SOA RR) with:

```
$ dnssec-signzone -f- test-data/example.rfc5155 test-data/Kexample.+008+31967
dnssec-signzone: error: dns_master_load: test-data/example.rfc5155:2: example: not at top of zone
dnssec-signzone: fatal: failed loading zone from 'test-data/example.rfc5155': not at top of zone
```

Specifying `-o` to `dnssec-signzone` fixes that problem:

```
$ dnssec-signzone -o example. -f- test-data/example.rfc5155 test-data/Kexample.+008+31967
; File written on Wed May 21 12:11:48 2025
; dnssec_signzone version 9.18.33
example.                                      3600 IN SOA       ns1.example. bugs.x.w.example. 1 3600 300 3600000 3600
...
```

ximon18: After some internal discussion I changed this so in `dnst` mode `-o` is mandatory, and in `ldns` mode that `-o` is optional with detection of the first SOA as the apex.

This isn't great, but is safest for DNST and most compatible for LDNS.

We expect to revisit the CLI interface anyway so for now maybe this is good enough, if odd to have a mandatory option (but as said above `dnsssec-signzone` seems to work that way too).

PR #68, opened 2025-05-13
Support signing of zones without `$ORIGIN` or `-o`.

ximon18:
Depends on `domain` PR https://github.com/NLnetLabs/domain/pull/527.

With related unit and integration test that deliberately do not pass the `-o` argument (to specify the origin), instead requiring the zone parser to determine the origin from the initial `example.` owner, as `ldns-signzone` does. 

This PR merges into PR #8.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -337,7 +337,7 @@ dependencies = [
 [[package]]
 name = "domain"
 version = "0.10.3"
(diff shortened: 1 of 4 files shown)


ximon18:
I will abort this PR per the final comment on https://github.com/NLnetLabs/domain/pull/527.

PR #72, opened 2025-05-15
Add a test of '-u'.

ximon18:
That sets the SOA SERIAL to the Epoch time now if ahead of the SOA SERIAL, or increments the SOA SERIAL otherwise.

Changes the code that determines the serial now to do so with its own mockable source of time from the environment, instead of using `Serial::now()` which hard-codes use of the real system clock, and extends the concept of `Env` to support seconds since the epoch which can be overridden when using `FakeEnv`.

### DIFF
--- a/src/env/fake.rs
+++ b/src/env/fake.rs
@@ -194,7 +194,7 @@ impl FakeCmd {
                 .map(|s| (s, Arc::new(CurrStepValue::new()))),
         };
 
-        (env_modifier)(&mut env);
+        env_modifier(&mut env);
 
         let exit_code = run(&mut env);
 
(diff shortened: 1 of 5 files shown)


src/env/fake.rs
```diff
@@ -157,24 +171,32 @@ impl FakeCmd {
...
             stderr: Default::default(),
+            seconds_since_epoch: None,
             stelline: self
                 .stelline
                 .clone()
                 .map(|s| (s, Arc::new(CurrStepValue::new()))),
         };
 
-        let exit_code = run(&env);
+        (env_modifier)(&mut env);
```
Philip-NLnetLabs, line 197:
```suggestion
        env_modifier(&mut env);
```

Philip-NLnetLabs (approved):

PR #76, opened 2025-05-15
Add test for `-M`.

ximon18:
This PR merges into PR #8.


Philip-NLnetLabs (changes requested):
Instead of adding a test case it seems to me that it would be better to remove the -M flag. This flag does not exist in ldns and it is not clear if it is needed.

ximon18:
Closing as we should limit the `dnst` interface to just those commands we know we need at present to avoid needing to preserve unnecessary functionality in any future improvements we decide to make.

PR #80, opened 2025-05-22
Switch to a release version of domain.

ximon18:
While this PR passes CI checks, it won't yet work for the `packaging` workflow as the Docker build will fail is it depends on not-yet-merged `domain` PR https://github.com/NLnetLabs/domain/pull/529.


ximon18:
This PR is outdated, the switch to a release version of domain was already done in a separate PR.

PR #85, opened 2025-05-29
Add support for -I to notify.

ximon18:
Note: Rather than adding `SpecificIpUdpConnect` as this PR does it might be better to add a setter to `UdpConnect` in domain called `set_src(addr: SocketAddr)` and have it use the provided value, if any, else fallback to a v4 or v6 zero address otherwise.

Otherwise, a better name for `SpecificIpUdpConnect` might be `UdpConnectWithSourceAddr` or `UdpConnectFrom`.

### DIFF
--- a/doc/manual/source/man/dnst-notify.rst
+++ b/doc/manual/source/man/dnst-notify.rst
@@ -23,6 +23,10 @@ Options
 
       The zone to send the NOTIFY for.
 
+.. option:: -I <ADDRESS>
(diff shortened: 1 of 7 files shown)


tertsdiepraam (approved):
Great! Glad you figured it out!

ximon18:
While I think the API could be neater, especially if this were handled in `domain`, this will do for now.

PR #91, opened 2025-06-03
installation instructions

AlexanderBand:


doc/manual/source/building.rst
```diff
@@ -0,0 +1,113 @@
+Building From Source
+====================
+
+In order to build dnst from the source, you need to have Rust installed on
+your system. The Rust compiler runs on, and compiles to, a great number of
+platforms, though not all of them are equally supported. The official `Rust
+Platform Support`_ page provides an overview of the various support levels.
+
+While some system distributions include Rust as system packages, Routinator
```
ximon18, line 9: Probably you meant dnst here, not Routinator.

doc/manual/source/building.rst
```diff
@@ -0,0 +1,113 @@
...
+Cargo fetches and builds dnst’s dependencies into an executable binary
+for your platform. By default you install from crates.io, but you can for
+example also install from a specific Git URL, as explained below.
+
+Installing the latest dnst release from crates.io is as simple as
+running:
+
+.. code-block:: text
+
+  cargo install --locked dnst
```
ximon18, line 43: This won't work for now because you'd need `--git`. This is because we can't publish dnst to crates.io until another domain release is made first.

doc/manual/source/building.rst
```diff
@@ -0,0 +1,126 @@
+Building From Source
```
ximon18, line 1: Do we need a note saying that dnst is not yet published to crates.io so only building with `--git` is possible?

ximon18:
The Windows CI failure can be ignored, it has nothing to do with the changes in this PR.

PR #97, opened 2025-06-18
Remove From<io::Error> impl for Error.

Philip-NLnetLabs:
Remove From<io::Error> impl for Error because context tends to get lost.

### DIFF
--- a/src/commands/signzone.rs
+++ b/src/commands/signzone.rs
@@ -933,7 +933,9 @@ impl SignZone {
         let mut writer = if out_file.as_os_str() == "-" {
             FileOrStdout::Stdout(env.stdout())
         } else {
-            let file = File::create(env.in_cwd(&out_file))?;
+            let file = File::create(env.in_cwd(&out_file)).map_err::<Error, _>(|e| {
+                format!("unable to create file {}: {e}", out_file.display()).into()
+            })?;
             let file = BufWriter::new(file);
             FileOrStdout::File(file)
         };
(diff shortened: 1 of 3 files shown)


Philip-NLnetLabs:
One thing to note. There a .context method. It is not used in update and sometimes in signzone. Does it add anything?

mozzieongit:
It adds a `... while {context}` string to the error message, e.g (not actually there, but could be in update):
```
Error: malformed IP address: <something-not-a-ip>
... while parsing rdata argument for AAAA
```
or
```
Error: connection timed out
... while fetching nameservers for zone {}
```

We should use the context method more often, but it's easy to forget.

src/commands/signzone.rs
```diff
@@ -1360,14 +1362,29 @@ impl SignZone {
...
-        let zone_file_len = zone_file.metadata()?.len();
+        let zone_file_len = zone_file
+            .metadata()
+            .map_err::<Error, _>(|e| {
+                format!(
+                    "error getting metadata from zonefile {}: {e}",
+                    zonefile_path.display()
+                )
+                .into()
+            })?
```
tertsdiepraam, line 1378: I think this should work too (removing both the type annotation and the `into`)
```suggestion
            .map_err(|e| {
                format!(
                    "error getting metadata from zonefile {}: {e}",
                    zonefile_path.display()
                )
            })?
```

tertsdiepraam (approved):

PR #99, opened 2025-06-23
Add support for the KMIP cryptographic backend.

ximon18:
Uses the `patches-for-nameshed-prototype` branch of `domain` which includes all (?) of the changes in the `crypto-and-keyset-fixes`.

Changes include:
- Upgrade domain to get KMIP support.
- Add `kmip` subcommands to enable KMIP servers to be configured for use.
- Extend `fn new_keys()` to support KMIP.
- Extend `fn update_dnskey_rrset()` to support MKIP.
- Factor out key removal into new `fn remove_key()` that also supports KMIP.
- Handle potential failure of DNSKEY lookup (which can heppen with KMIP keys).
- Including the DNS key tag in the generated KMIP key label, along with optional custom prefix, and DNS label based name truncation to fit within a specified maximum label length (default 32 bytes as this appears to be the lowest common value in use by KMIP/PKCS#11 implementations based on a brief Internet search).

Example usage:

```
$ dnst keyset -c /tmp/my.conf create -n example.com -s /tmp/my.state

$ dnst keyset -c /tmp/my.conf kmip add-server my_server 127.0.0.1 --credential-store /tmp/my.creds --insecure --username Nameshed --password 1234 --port 5696

$ dnst keyset -c /tmp/my.conf kmip list-servers
Servers:
    ID: my_server [DEFAULT]
        Address:                           127.0.0.1:5696
        Server Certificate Verification:   Disabled
        Server Certificate:                None
        Certificate Authority Certificate: None
        Client Credentials:                /tmp/x.creds
        Client Certificate Authentication: Disabled
        Client Limits:
            Connect Timeout:               10 seconds
            Read Timeout:                  10 seconds
            Write Timeout:                 10 seconds
            Max Response Size:             8192 bytes

$ cat /tmp/x.creds 
{
  "my_server": {
    "username": "Nameshed",
    "password": "1234"
  }

$ dnst keyset -c /tmp/my.conf kmip remove-server my_server
[target/debug/dnst] ERROR: KMIP server 'my_server' cannot be removed as it is the current default. Use kmip disable first.

$ dnst keyset -c /tmp/my.conf kmip disable

$ dnst keyset -c /tmp/my.conf kmip remove-server my_server

$ cat /tmp/x.creds 
{}

$ dnst keyset -c /tmp/my.conf kmip list-servers
Servers:
```

Now with actual key generation:

```
$ dnst keyset -c /tmp/my.conf kmip list-servers
Servers:
    ID: my_server [DEFAULT]
...

$ dnst keyset -c /tmp/my.conf init
Get DNSKEY RRset: [...]

$ dnst keyset -c /tmp/my.conf kmip keys
Keys:
	kmip://my_server/keys/8102A300E886331D9D6026759FBEE7AD1EB8D54A_pub?algorithm=8&flags=257 kmip://my_server/keys/8102A300E886331D9D6026759FBEE7AD1EB8D54A_priv?algorithm=8&flags=257
		Type: KSK, algorithm: 8, key tag: 28211
		State: Signer, Present (Active)
		Created: 2025-08-19T14:35:19
		Published: 2025-08-19T14:35:19
		Visible: <empty>
		DS visible: <empty>
		RRSIG visible: <empty>
		Withdrawn: <empty>
	kmip://my_server/keys/0CF865359C44BC8FF0F520996C8F161D59C8581E_pub?algorithm=8&flags=256 kmip://my_server/keys/0CF865359C44BC8FF0F520996C8F161D59C8581E_priv?algorithm=8&flags=256
		Type: ZSK, algorithm: 8, key tag: 51745
		State: Signer, Present (Active)
		Created: 2025-08-19T14:35:19
		Published: 2025-08-19T14:35:19
		Visible: <empty>
		DS visible: <empty>
		RRSIG visible: <empty>
		Withdrawn: <empty>

$ dnst keyset -c /tmp/my.conf kmip remove-server my_server
[target/debug/dnst] ERROR: KMIP server 'my_server' cannot be removed as it is the current default. Use kmip disable first.

$ dnst keyset -c /tmp/my.conf kmip disable

$ dnst keyset -c /tmp/my.conf kmip remove-server my_server
[target/debug/dnst] ERROR: KMIP server 'my_server' cannot be removed as there are still keys using it.
```

### DIFF
--- a/src/commands/keyset/cmd.rs
+++ b/src/commands/keyset/cmd.rs
@@ -1375,6 +1375,7 @@ pub struct KeySetState {
 
     /// KMIP related configuration.
     #[cfg(feature = "kmip")]
+    #[serde(default)]
     pub kmip: KmipState,
 }
 


Philip-NLnetLabs (changes requested):
Looks good, though we should not store password in config file.

src/commands/keyset.rs
```diff
@@ -1323,6 +1474,144 @@ struct KeySetConfig {
 
     /// Automatically remove keys that are no long in use.
     autoremove: bool,
+
+    /// KMIP servers to use, keyed by user chosen HSM id.
+    kmip_servers: HashMap<String, KmipServerConnectionSettings>,
```
Philip-NLnetLabs, line 1479: I wonder if this should move to KeySetState. Signers read KeySetState to figure out which keys to sign with. I think it would be nicer if all information is there. Otherwise we have to have a standard file format to KeySetConfig as well.

ximon18:
> Looks good, though we should not store password in config file.

Maybe it should also use a Secret type or such when in-memory to prevent leakage?

src/commands/keyset.rs
ximon18: @Philip-NLnetLabs and I discussed this and agreed to move the KMIP config into the `KeyStateState` because that is all that "downstreams" such as `dnst signer` have available to them.

ximon18:
> Should the key naming pattern perhaps even be a user controllable pattern?

Some further thoughts on this:

Use an overridable naming template? E.g. support placeholders such as `<key type>`, `<uniqifier>` and `<zone name>`, with a default of:

```
  <zone name>-<uniqifier>-<key type>
```

Where `<uniqifier>` is 2 bytes long and `<key type>` is 3 bytes long, leaving 32 - '-' - 2 - '-' - 3 = 32 - 7 = 25 bytes for `<zone name>`, which can be abbreviated if too long by replacing the middle with '...' and `<uniqifier>` is a 0 padded positive integer in the range 00..99 giving 100 keys to roll the zone up to twice a week without needing to use 00 again.

When overridden a user could include fixed namespace and application values, e.g.:

```
  NS-PROD-<zone name>-<uniqifier>-<key type>
```

Resulting in key names like:

```
  NS-PROD-example.com-001-ksk
  NS-DEV-some.lo...in-name-013-zsk
  (shrunk from NS-DEV-some.long-domain-name-013-zsk)
  01234567890123456789012345678901
```

However, regarding `<uniqifier>`, it may be that pre-generation should be accomplished differently, by generating the keys outside of `dnst keyset` and importing them. But it may still be useful to consider what to do if a key fails to generate, should we retry with an integer value at the end of the zone name (within the 32 byte limit - aside: should that limit also
be configurable?), can we even tell that failure was due to a name collision?

Or is this all overkill?

Philip-NLnetLabs (approved):

PR #100, opened 2025-07-08
Clapify keyset

Philip-NLnetLabs:
Use clap to parse subcommands of keyset.

### DIFF
--- a/src/commands/keyset.rs
+++ b/src/commands/keyset.rs
@@ -39,20 +39,6 @@ pub struct Keyset {
     #[arg(short = 'c')]
     keyset_conf: PathBuf,
 
-    /// Domain name (only for create)
-    #[arg(short = 'n')]
-    domain_name: Option<Name<Vec<u8>>>,
-
-    /// State file (only for create)
-    #[arg(short = 's')]
-    keyset_state: Option<PathBuf>,
-
-    #[arg(short = 't')]
-    ttl: Option<u32>,
-
-    #[arg(short = 'b')]
-    bits: Option<usize>,
-
     /// Subcommand
     #[command(subcommand)]
     cmd: Commands,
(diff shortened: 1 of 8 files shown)


src/commands/keyset.rs
```diff
@@ -38,25 +39,146 @@ pub struct Keyset {
...
-    cmd: String,
+    #[command(subcommand)]
+    cmd: Commands,
+}
 
-    value: Option<String>,
+type OptDuration = Option<Duration>;
+
+#[derive(Clone, Debug, Subcommand)]
+enum Commands {
```
tertsdiepraam, line 50: Would it make sense to split this up into subcommands? For example make a `set` and `get` subcommand. I feel like that would make it easier to navigate.

src/commands/keyset.rs
```diff
@@ -155,329 +274,133 @@ impl Keyset {
...
-                        })?;
-                    } else {
-                        panic!("unsupported URL scheme in {ksk_pub_url}");
-                    }
-
-                    return Err(e);
+        match self.cmd {
+            Commands::Create {
+                domain_name: _,
+                keyset_state: _,
```
tertsdiepraam, line 280:
```suggestion
                ..
```

Small nit

tertsdiepraam (commented):
Happy 100th PR on dnst 🎉

src/commands/keyset.rs
Philip-NLnetLabs: Yes. I'll take a look what that leads to.

src/commands/keyset.rs
```diff
@@ -1100,6 +1175,128 @@ impl Keyset {
...
+
+fn set_command(
+    cmd: SetCommands,
+    ksc: &mut KeySetConfig,
+    config_changed: &mut bool,
+) -> Result<(), Error> {
+    match cmd {
+        SetCommands::UseCsk { boolean } => {
+            ksc.use_csk = boolean;
+            *config_changed = true;
```
ximon18, line 1238: Since every match block sets config_changed = true, maybe move that out to the end?

src/commands/keyset.rs
Philip-NLnetLabs: This needs careful design also with feedback from users. Keyset is mostly incompatible with the concepts in Yuri's paper, so we only try to explain that in the documentation. But it is indeed out of scope for this PR.

src/commands/keyset.rs
```diff
@@ -1740,6 +1937,7 @@ fn handle_actions(
             Action::ReportDsPropagated => (),
             Action::ReportRrsigPropagated => (),
             Action::WaitDnskeyPropagated => (),
+            Action::WaitDsPropagated => (),
```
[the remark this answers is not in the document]
Philip-NLnetLabs: It is a change to domain that is reflected here.

PR #103, opened 2025-08-04
Improve error message when failing to print to stdout

### LINKED ISSUE #49: Confusing error (panic) when writing to a broken stdout/pipe
While accidentally piping dnst output into an already exited program (it failed due to invalid arguments), dnst panics, as https://github.com/NLnetLabs/dnst/blob/6f0e27d3e15547818466152790787122e644432b/src/env/mod.rs#L71 uses unwrap on the output stream.
This leads to an ugly error that was confusing at first.
```
$ target/release/dnst nsec3-hash hi | sort -k1s
sort: stray character in field spec: invalid field specification ‘1s’
thread 'main' panicked at /home/mozzie/repos/work/dnst.trees/main/src/env/mod.rs:71:32:
called `Result::unwrap()` on an `Err` value: Error
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

Maybe this should change to either:
1. an attempt to write to stderr that writing to stdout failed (if that also fails, ignore and proceed to exit), and exit with an error code; or
2. just exit with an error code

mozzieongit:
Fixes #49

### DIFF
--- a/src/env/mod.rs
+++ b/src/env/mod.rs
@@ -101,7 +101,8 @@ impl<T: io::Write> Stream<T> {
             // terminates before we do. Note: If stderr is also broken this
             // will panc, but then neither the panic nor this error would be
             // printed out.
-            eprintln!("dnst: failed printing to stdout: {e}");
+            let prog = std::env::args_os().next().unwrap().to_string_lossy().to_string();
+            eprintln!("{prog}: failed printing to stdout: {e}");
             // using exit(101) to replicate rust's default error exit code
             std::process::exit(101);
         }
--- a/src/env/mod.rs
+++ b/src/env/mod.rs
@@ -97,8 +97,10 @@ impl<T: io::Write> io::Write for &Stream<T> {
 impl<T: io::Write> Stream<T> {
     pub fn write_fmt(&mut self, args: fmt::Arguments<'_>) {
         if let Err(e) = self.writer.lock().unwrap().deref_mut().write_fmt(args) {
-            // This will panic, if stderr is also broken, but then neither the
-            // panic nor this error would be printed out.
+            // This can happen when output is piped to a command that
+            // terminates before we do. Note: If stderr is also broken this
+            // will panc, but then neither the panic nor this error would be
+            // printed out.
             eprintln!("dnst: failed printing to stdout: {e}");
             // using exit(101) to replicate rust's default error exit code
             std::process::exit(101);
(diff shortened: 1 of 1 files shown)


ximon18:
Thanks for looking at this @mozzieongit.

One minor observation:

```
$ cargo run -q -- nsec3-hash hi | sort -k1s
sort: stray character in field spec: invalid field specification ‘1s’
failed printing to stdout: Broken pipe (os error 32)
```

As a user, how do I know that `failed printing...` came from `dnst` ? For the `sort` command output it is prefixed by `sort: `.

I don't think we should block this PR on this point, nor am I even sure what we should do about it ... If we had a richer log style output it might be clearer where the output came from but that could be be considered to be too detailed compared to typical stderr output...

src/env/mod.rs
```diff
@@ -96,17 +96,13 @@ impl<T: io::Write> io::Write for &Stream<T> {
...
-            .unwrap()
-            .deref_mut()
-            .write_fmt(args)
-            .unwrap();
+        if let Err(e) = self.writer.lock().unwrap().deref_mut().write_fmt(args) {
+            // This will panic, if stderr is also broken, but then neither the
+            // panic nor this error would be printed out.
+            eprintln!("failed printing to stdout: {e}");
+            // using exit(101) to replicate rust's default error exit code
+            std::process::exit(101);
```
ximon18, line 104: Hmm, I'm not familiar with exit code 101, I didn't realize Rust did that. Googling I found https://users.rust-lang.org/t/solved-why-101-exit-code-when-use-panic/80061/4 which points to a commit which says that on *nix `libc::abort()` is used but am no clearer as to why that results in exit code 101. Anyway, the question on my mind is do we want a Rust particular behaviour, or would it be better to just exit with our normal exit code 1 (see `error::exit_code()`).

mozzieongit: I found the exit code in the rustc code.

Either is fine by me. We're currently not using exit codes to convey error types anyway.

ximon18: Let's go with what you chose to do, as it maintains existing behaviour and you left a comment explaining what it is doing and why.

ximon18:
> > I don't think we should block this PR on this point, nor am I even sure what we should do about it ... If we had a richer log style output it might be clearer where the output came from but that could be be considered to be too detailed compared to typical stderr output...
> 
> I'd say for a start we just put `dnst:` in front. And if that error comes from the ldns-compat tool, so be it. If people get confused by what dnst is when they're using the compat tools, I hope they'll look at their installed packages and see that dnst is the successor to the ldns-tool they're using at that moment.

I guess ideally thinking about it this would print the argv[0] name instead of hard-coded `dnst`, but if that's overkill / not easy to do at this point then ignore me...

ximon18 (approved):

Issue #90, opened 2025-06-03
ldns-signzone emulation orders output RRs differently to original ldns-signzone.


ximon18:
See https://github.com/NLnetLabs/dnst/issues/43#issuecomment-2934045552

I think this is because we chose to prefer speed by default over ordering. For `dnst signzone` there is a `-R` argument that does the ordering if wanted.

I think the `-R` argument also existed for `ldns-signzone` emulation but I removed it thinking it wasn't part of the original ldns-signzone so should not be supported/advertised.

Is this a bug?

Should we default `-R` to true for `ldns-signzone` mode?

mozzieongit:
> Is this a bug?

Maybe. Depends on how exact we want to be with the compatibility mode. But considering we changed the date output to match ldns, we might as well match the ordering.
 
> Should we default `-R` to true for `ldns-signzone` mode?

Seems reasonable and simple enough. I'll make a PR.

Issue #65, opened 2025-05-02
key2ds: Make hex output lowercase to match the original ldns.


ximon18:
The `key2ds` subcommand currently outputs uppercase hex while the original `ldns-key2ds` uses lowercase hex:

E.g.

```
$ cargo run -q -- key2ds -n /tmp/keys/Kexample.com.+013+21792.key 
example.com.    3600    IN      DS      21792 13 2 EF8FA37064BDA8BBAD6829F343A25156D57F19ED4E7B02CAD2157E11FC855123
```

Compared to:

```
$ ldns-key2ds -n /tmp/keys/Kexample.com.+013+21792.key 
example.com.    3600    IN      DS      21792 13 2 ef8fa37064bda8bbad6829f343a25156d57f19ed4e7b02cad2157e11fc855123
```

PR #109, opened 2025-08-06
Match ldns-signzone NSEC/NSEC3 RR ordering and NSEC3 formatting

mozzieongit:
Fixes: #90

---

I repeatedly ask myself, how exact do we want to match ldns-signzone formatting. Do we copy every weird quirk or do we fix them? e.g. the weird white space in the NSEC3 records. 
Considering @ximon18 asked to make the DS digest lowercase in #65, I assumed you'd note the same about the NSEC3 hash and changed that (and added the weird white space while I'm at it).
Copying every weird quirk makes it easier to test and match against ldns-signzone output of course.

Also, I feel like the signzone code has become quite the beast and I'd propose splitting it into ldns-compat-signzone and dnst-signzone functions (in some future), where we can easily fix things in the dnst version without worrying about ldns-signzone compatibility, and can improve other things as well.

### DIFF
--- a/src/commands/signzone.rs
+++ b/src/commands/signzone.rs
@@ -1264,7 +1264,21 @@ impl SignZone {
                             && rrset.owner() == apex)
                 }) {
                     for rr in rrset.iter() {
-                        self.write_rr(&mut writer, rr)?;
+                        if self.invoked_as_ldns {
+                            if let ZoneRecordData::Nsec3(nsec3) = rr.data() {
+                                let rr =
+                                    Record::new(rr.owner(), rr.class(), rr.ttl(), LdnsNsec3(nsec3));
+                                writer.write_fmt(format_args!(
+                                    "{}",
+                                    rr.display_zonefile(DISPLAY_KIND)
+                                ))?;
+                            } else {
+                                self.write_rr(&mut writer, rr)?;
+                            }
+                        } else {
+                            self.write_rr(&mut writer, rr)?;
+                        }
+
                         match rr.data() {
                             ZoneRecordData::Nsec3(nsec3) if self.extra_comments => {
                                 nsec3.comment(&mut writer, rr, nsec3_cs)?
@@ -1282,6 +1296,7 @@ impl SignZone {
                         }
                         writer.write_str("\n")?;
                     }
+
                     // Now attempt to print the RRSIGs that covers the RTYPE of this RRSET.
                     for covering_rrsigs in owner_rrs
                         .rrsets()
@@ -2074,6 +2089,52 @@ impl<O: AsRef<[u8]>, N: ToName> ZonefileFmt for YyyyMmDdHhMMSsRrsig<'_, O, N> {
 impl<O, N> RecordData for YyyyMmDdHhMMSsRrsig<'_, O, N> {
     fn rtype(&self) -> Rtype {
         Rtype::RRSIG
+    }
+}
+
+//------------ LdnsNsec3 -----------------------------------------------------
+
+/// A wrapper around Nsec3 to print the Nsec3 data in the exact format used by
+/// ldns-signzone with all its quirks.
+struct LdnsNsec3<'a, O>(&'a Nsec3<O>);
+
+impl<O: AsRef<[u8]>> ZonefileFmt for LdnsNsec3<'_, O> {
+    fn fmt(&self, p: &mut impl Formatter) -> zonefile_fmt::Result {
+        // This block of code was copied from the `domain` crate impl of
+        // `Zonefilefmt` for domain::rdata::nsec3::Nsec3 and adapted for
+        // ldns output format.
+        p.block(|p| {
+            p.write_show(self.0.hash_algorithm())?;
+            p.write_token(self.0.flags())?;
+            p.write_comment(format_args!(
+                "flags: {}",
+                if self.0.opt_out() {
+                    "opt-out"
+                } else {
+                    "<none>"
+                }
+            ))?;
+            p.write_token(self.0.iterations())?;
+            p.write_comment("iterations")?;
+            p.write_show(self.0.salt())?;
+            p.write_token(format!(
+                " {}",
+                domain::utils::base32::encode_display_hex(&self.0.next_owner())
+                    .to_string()
+                    .to_lowercase()
+            ))?;
+            p.write_show(self.0.types())?;
+            // ldns-signzone ends its NSEC3 rtype bitmap with a trailing
+            // space. Adding an empty token, because the formatter will add
+            // a space as a delimiter.
+            p.write_token("")
+        })
+    }
+}
+
+impl<O> RecordData for LdnsNsec3<'_, O> {
+    fn rtype(&self) -> Rtype {
+        Rtype::NSEC3
     }
 }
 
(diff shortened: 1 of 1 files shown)


ximon18:
Replying, with slight re-ordering of your original text:

> I repeatedly ask myself, how exact do we want to match ldns-signzone formatting. Do we copy every weird quirk or do we fix them? e.g. the weird white space in the NSEC3 records.
> Copying every weird quirk makes it easier to test and match against ldns-signzone output of course.

In LDNS emulation mode, as closely as possible both to match user expectations (though we may be exceeding those) and as you say to make comparison easier for diagnostic purposes.

> Considering @ximon18 asked to make the DS digest lowercase in #65, I assumed you'd note the same about the NSEC3 hash and changed that (and added the weird white space while I'm at it).

👍 

> Also, I feel like the signzone code has become quite the beast and I'd propose splitting it into ldns-compat-signzone and dnst-signzone functions (in some future), where we can easily fix things in the dnst version without worrying about ldns-signzone compatibility, and can improve other things as well.

Sounds good, I think it's partly a question of time, resources and priorities, not that anyone would object to such a change (at least I would not object 😉 ).

src/commands/signzone.rs
```diff
@@ -2967,14 +3072,14 @@ m.root-servers.net.\t3600000\tIN\tAAAA\t2001:dc3::35
...
         let ldns_dnst_output_stripped: &str = "\
-            example.org.\t3600\tIN\tRRSIG\tNSEC3PARAM 15 2 3600 20241127162422 20241127162422 38873 example.org. 0XdDm1l2Mm8dyhtzbyQb91CmyNONs8lc9d22FUGvpjfqo8T2h0xs04x5MIfP0DjmiVnNqIyPK6sipnDqf6tCDg==\n\
             example.org.\t3600\tIN\tNSEC3PARAM\t1 1 1 -\n\
+            example.org.\t3600\tIN\tRRSIG\tNSEC3PARAM 15 2 3600 20241127162422 20241127162422 38873 example.org. 0XdDm1l2Mm8dyhtzbyQb91CmyNONs8lc9d22FUGvpjfqo8T2h0xs04x5MIfP0DjmiVnNqIyPK6sipnDqf6tCDg==\n\
+            93u63bg57ppj6649al2n31l92iedkjd6.example.org.\t240\tIN\tNSEC3\t1 1 1 -  k71ku6aicr5jpdjoe9j7cdnlk6d5c3ue A NS SOA RRSIG DNSKEY NSEC3PARAM \n\
             93u63bg57ppj6649al2n31l92iedkjd6.example.org.\t240\tIN\tRRSIG\tNSEC3 15 3 240 20241127162422 20241127162422 38873 example.org. z4ceUmbSZiSnluFj8CDJ7B9fukCR2flTWgca4GE2xrw48+fiieH/04xCKhJmDRJUJTVkKtIYpB4p0Q4m60M1Cg==\n\
-            93u63bg57ppj6649al2n31l92iedkjd6.example.org.\t240\tIN\tNSEC3\t1 1 1 - K71KU6AICR5JPDJOE9J7CDNLK6D5C3UE A NS SOA RRSIG DNSKEY NSEC3PARAM\n\
+            k71ku6aicr5jpdjoe9j7cdnlk6d5c3ue.example.org.\t240\tIN\tNSEC3\t1 1 1 -  ojicmhri4vp8po7h2kvej99sklqnj5p2 NS \n\
```
ximon18, line 3079: This RR matches the original one that was output further down, but what happened to the `93u6.. -> k71k` NSEC3 chain RR that covered `A NS SOA RRSIG DNSKEY NSEC3PARAM`, that appears to have been replaced by an RR from `ojic.. -> 93u6..` covering only `NS DS RRSIG`? I wouldn't expect an output ordering / formatting change to affect the NSEC3 chain.

src/commands/signzone.rs
```diff
@@ -1255,12 +1256,29 @@ impl SignZone {
             // we skip that, and we skip RRSIGs as they are output only after
             // the RRset that they cover.
             if self.order_rrsigs_after_the_rtype_they_cover {
-                for rrset in owner_rrs
-                    .rrsets()
-                    .filter(|rrset| !matches!(rrset.rtype(), Rtype::SOA | Rtype::RRSIG))
-                {
+                for rrset in owner_rrs.rrsets().filter(|rrset| {
+                    !(matches!(rrset.rtype(), Rtype::SOA | Rtype::RRSIG)
+                        || self.invoked_as_ldns
```
ximon18, line 1261: This extension to the condition could do with an explanatory comment, and I had to lookup the precedence of || and && to understand how the logic would be evaluated so for me at least some parentheses would help too.

I don't yet know what your change here is doing or why.

src/commands/signzone.rs
```diff
@@ -1283,13 +1301,54 @@ impl SignZone {
                     for covering_rrsigs in owner_rrs
                         .rrsets()
                         .filter(|this_rrset| this_rrset.rtype() == Rtype::RRSIG)
-                        .map(|this_rrset| this_rrset.iter().filter(|rr| matches!(rr.data(), ZoneRecordData::Rrsig(rrsig) if rrsig.type_covered() == rrset.rtype())))
+                        .map(|this_rrset| {
+                            this_rrset.iter().filter(|rr| {
+                                matches!(rr.data(), ZoneRecordData::Rrsig(rrsig)
+                                    if rrsig.type_covered() == rrset.rtype()
+                                        // && (!self.invoked_as_ldns || rr.owner() != apex
```
ximon18, line 1308: Should this commented out code be removed?

src/commands/signzone.rs
```diff
@@ -1283,13 +1301,54 @@ impl SignZone {
...
+                            })
+                        })
+                    {
+                        for covering_rrsig_rr in covering_rrsigs {
+                            self.writeln_rr(&mut writer, covering_rrsig_rr)?;
+                        }
+                    }
+                }
+
+                if self.invoked_as_ldns && !self.use_nsec3 && owner_rrs.owner() == apex {
```
ximon18, line 1323: This combination of testing for invoked as LDNS, not using NSEC3 and the owner being the apex is done in several places in the added code. Perhaps it could be factored out to a helpfully named variable? I would suggest a name but I'm still not sure what you are doing here or why 😉 ...

src/commands/signzone.rs
```diff
@@ -1255,12 +1256,29 @@ impl SignZone {
...
+                        || self.invoked_as_ldns
+                            && !self.use_nsec3
+                            && rrset.rtype() == Rtype::NSEC
+                            && rrset.owner() == apex)
+                }) {
                     for rr in rrset.iter() {
-                        self.write_rr(&mut writer, rr)?;
+                        if self.invoked_as_ldns {
+                            if let ZoneRecordData::Nsec3(nsec3) = rr.data() {
+                                let rr =
```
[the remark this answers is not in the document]
mozzieongit: Leftover from testing. Moving it into the function.

src/commands/signzone.rs
mozzieongit: Hopefully the new comments clear things up.

src/commands/signzone.rs
```diff
@@ -1248,12 +1249,23 @@ impl SignZone {
...
+                    !(matches!(rrset.rtype(), Rtype::SOA | Rtype::RRSIG)
+                    // If run as ldns-signzone we want to list the NSEC RR
+                    // at the end of the RRset of the apex. By default,
+                    // the NSEC RR would preceed the DNSKEY RRset, so we
+                    // need to filter it out here to manually reinsert it
+                    // later. This is only necessary for the NSEC RR at
+                    // the apex, as the ordering issue doesn't appear at
+                    // other locations than the apex.
+                        || (self.invoked_as_ldns
+                            && !self.use_nsec3
```
ximon18, line 1262: `!self.use_nsec3` is redundant here as if it's using NSEC3 it won't match Rtype::NSEC, so you can remove this line.

src/commands/signzone.rs
```diff
@@ -1276,13 +1288,59 @@ impl SignZone {
...
+                if self.invoked_as_ldns && !self.use_nsec3 && owner_rrs.owner() == apex {
+                    for covering_nsec in owner_rrs
+                        .rrsets()
+                        .filter(|this_rrset| this_rrset.rtype() == Rtype::NSEC)
+                        .map(|this_rrset| this_rrset.iter())
+                    {
+                        for covering_nsec_rr in covering_nsec {
+                            self.writeln_rr(&mut writer, covering_nsec_rr)?;
+                        }
+                    }
```
ximon18, line 1324: This could be simplfied by using `find()` and `first()` instead of `filter()` as there is only ever one NSEC RR in an RRset, e.g.:

```suggestion
                    if let Some(nsec_rrset) = owner_rrs
                        .rrsets()
                        .find(|this_rrset| this_rrset.rtype() == Rtype::NSEC)
                    {
                        self.writeln_rr(&mut writer, nsec_rrset.first())?;
                    }
```

Issue #42, opened 2024-11-29
Improve `dnst update` user interface and functionality?


mozzieongit:
While reviewing #10 I [mentioned](https://github.com/NLnetLabs/dnst/pull/10#discussion_r1862371109) that:

1) we should improve on the interface a bit (the "none" in place of an IP address for deletion is bugging me in particular)
2) that it might be useful for `dnst update` to be able to update more than just A/AAAA records, like TXT records (e.g. for acme challenges).

An initial idea would be the following interface variants:

### Variant A

Synopsis:
- `dnst update [options] add <domain name> <RRTYPE> <RRs>...`
- `dnst update [options] delete <domain name> <RRTYPE> [<RRs>...]`
- `dnst update [options] clear <domain name>` (this distinction is to avoid accidental deletion of whole domain names, but might be unnecessary)

Examples:
- `dnst update add <domain name> AAAA "::1" "::2"`
  - Add multiple AAAA records
- `dnst update add <domain name> TXT "challenge"`
  - Add multiple TXT records
- `dnst update delete <domain name> AAAA ::1 ::2`
  - Delete exact AAAA RRs on domain name (`::1` and `::2` in this case)
- `dnst update delete <domain name> AAAA`
  - Delete all AAAA RRs on domain name
- `dnst update clear <domain name>`
  - Delete all RRSETs on domain name, aka delete the whole domain name

### Variant B

(like above, but as options for different RR types at once)

Alternatively provide RRs as options, to be able to update multiple different RR types at once:

- `dnst update add <domain name> --a 1.1.1.1 --aaaa ::1 --txt "hello world"`
  - Add all these records with the TTL determined from the SOA RR
- `dnst update delete <domain name> --a 1.1.1.1 --aaaa ::1 ...`
  - Delete these exact records if they exist
- `dnst update delete <domain name> --a --aaaa ...`
  - Delete all mentioned RRsets on domain name
- `dnst update clear <domain name>`
  - Delete all RRSETs on domain name, aka delete the whole domain name

---

The update `[options]` could then set the --ttl and maybe even --class (but that is probably rarely helpful).

---

I like variant (A) more. And I think it's easier to implement generically over RRtype than the `--XYZ` option where we would have to specify an argument for every supported RRtype (could of course be made easy with a macro).

PR #115, opened 2025-08-19
Reimplement `dnst update` with better UI

mozzieongit:
The new syntax:
- `dnst update [options] <domain name> add <RRTYPE> <RDATA>...`
- `dnst update [options] <domain name> delete <RRTYPE> [<RDATA>...]`
- `dnst update [options] <domain name> clear` (distinction to avoid accidental deletion of whole domain names)

Options:
```
  -c, --class <CLASS>
          Class [default: IN]
  -t, --ttl <TTL>
          TTL in seconds or with unit suffix (s, m, h, d, w, M, y) [default: 3600]
  -s, --server <IP>
          Name server to send the update to (can be provided multiple times)
  -z, --zone <ZONE>
          Zone the domain name belongs to (to skip SOA query)
  -y, --tsig <NAME:KEY[:ALGO]>
          TSIG credentials for the UPDATE packet
      --rrset-exists <DOMAIN_NAME_AND_TYPE>
          RRset exists (value independent). (Optionally) provide this option multiple times, with format "<DOMAIN_NAME> <TYPE>" each, to build up a list of RR(set)s [aliases: --rrset]
      --rrset-exists-exact <RESOURCE_RECORD>
          RRset exists (value dependent). (Optionally) provide this option multiple times, each with one RR in zonefile format, to build up one or more RRsets that is required to exist. CLASS and TTL can be omitted [aliases: --rrset-exact]
      --rrset-non-existent <DOMAIN_NAME_AND_TYPE>
          RRset does not exist. (Optionally) provide this option multiple times, with format "<DOMAIN_NAME> <TYPE>" each, to build up a list of RRs that specify that no RRs with a specified NAME and TYPE can exist [aliases: --rrset-empty]
      --name-in-use <DOMAIN_NAME>
          Name is in use. (Optionally) provide this option multiple times, with format "<DOMAIN_NAME>" each, to collect a list of NAMEs that must own at least one RR [aliases: --name-used]
      --name-not-in-use <DOMAIN_NAME>
          Name is not in use. (Optionally) provide this option multiple times, with format "<DOMAIN_NAME>" each, to collect a list of NAMEs that must NOT own any RRs [aliases: --name-unused]
```

Examples:
- `dnst update <domain name> add AAAA "::1" "::2"` - Add multiple AAAA records
- `dnst update <domain name> add TXT "challenge1" '"Second RR with space" "and 2 strings"'` - Add multiple TXT records
- `dnst update <domain name> delete AAAA ::1 ::2` - Delete exact AAAA RRs on domain name (`::1` and `::2` in this case)
- `dnst update <domain name> delete AAAA` - Delete all AAAA RRs on domain name
- `dnst update <domain name> clear` - Delete all RRSETs on domain name, aka delete the whole domain name

I thought about adding the domain name to the action to have it behind the action command, but put it to the update command instead to reduce code duplication. But on the other hand it's not much duplication... 
So, any preferences on `dnst update example.com add TXT Hallo` vs. `dnst update add example.com TXT Hallo`?

Left over TODOs:

- [x] Check if more errors could use some `.context()`
- [x] "pass SOA record fetched above to make sure changes to the SOA record by the user adhere to the RFC specifications of e.g. only increasing serial"
  - The RFC requires updates to the SOA RR to only increase the serial.
- [ ] Add `from_rtype_and_str` to domain's ZoneRecordData to skip the workaround with zonefile?
  - Currently, I'm parsing the rdata using `inplace::Zonefile` by creating a temporary RR and using only the resulting parsed rdata.
- [ ] Add tests triggering the runtime checks
- [ ] Add stelline tests
- [x] Document the format of RDATA arguments (e.g. what to do with a TXT record using multiple quoted strings)
- [x] Update the dnst update man page

Fixes: #42

### DIFF
--- a/src/commands/mod.rs
+++ b/src/commands/mod.rs
@@ -74,6 +74,13 @@ pub enum Command {
     #[command(name = "update")]
     Update(self::update::Update),
 
+    /// Send an UPDATE packet ldns compatibility variant
+    ///
+    /// This variant is not a dnst command and only used to provide
+    /// a separate implementation from the dnst update command.
+    #[command(skip)]
+    LdnsUpdate(self::update::LdnsUpdate),
+
     /// Show the manual pages
     Help(self::help::Help),
 
@@ -94,6 +101,7 @@ impl Command {
             Self::Notify(notify) => notify.execute(env),
             Self::SignZone(signzone) => signzone.execute(env),
             Self::Update(update) => update.execute(env),
+            Self::LdnsUpdate(ldnsupdate) => ldnsupdate.execute(env),
             Self::Help(help) => help.execute(),
             Self::Report(s) => {
                 writeln!(env.stdout(), "{s}");
--- a/src/commands/update.rs
+++ b/src/commands/update.rs
@@ -1271,7 +1271,7 @@ mod test {
                     algorithm: Algorithm::Sha256,
                 }),
                 ttl: Ttl::from_secs(60),
-                nameservers: vec![[127, 0, 0, 9].into()],
+                nameservers: Some([127, 0, 0, 9].into()),
                 action: UpdateAction::Add {
                     rtype: Rtype::TXT,
                     rdata: vec!["Hallo".into()]
--- a/src/commands/update.rs
+++ b/src/commands/update.rs
@@ -411,7 +411,9 @@ impl Update {
         match zonefile.next_entry() {
             Ok(Some(Entry::Record(record))) => Ok(record.data().clone().flatten_into()),
             Ok(_) => unreachable!("We always create a record"),
-            Err(e) => Err(format!("Failed to parse rdata for {rtype} {rdata} -- Error: {e}").into()),
+            Err(e) => {
+                Err(format!("Failed to parse rdata for {rtype} {rdata} -- Error: {e}").into())
+            }
         }
     }
 
--- a/src/commands/update.rs
+++ b/src/commands/update.rs
@@ -36,31 +36,8 @@ use crate::Args;
 
 use super::{parse_os, parse_os_with, Command, LdnsCommand};
 
-// TODO: add .context() to errors?
-
-// UI:
-// Synopsis:
-// - `dnst update [options] add <domain name> <RRTYPE> <RRs>...`
-// - `dnst update [options] delete <domain name> <RRTYPE> [<RRs>...]`
-// - `dnst update [options] clear <domain name>` (distinction to avoid accidental deletion of whole domain names)
-
-// Examples:
-// - `dnst update add <domain name> AAAA "::1" "::2"` - Add multiple AAAA records
-// - `dnst update add <domain name> TXT "challenge"` - Add multiple TXT records
-// - `dnst update delete <domain name> AAAA ::1 ::2` - Delete exact AAAA RRs on domain name (`::1` and `::2` in this case)
-// - `dnst update delete <domain name> AAAA` - Delete all AAAA RRs on domain name
-// - `dnst update clear <domain name>` - Delete all RRSETs on domain name, aka delete the whole domain name
-
-// type SomeRecord = Record<NameVecU8, ZoneRecordData<Vec<u8>, NameVecU8>>;
-// type SomeRecord = Record<Name<Vec<u8>>, ZoneRecordData<Vec<u8>, Name<Vec<u8>>>>;
 type ParsedRecord = Record<Name<Vec<u8>>, ZoneRecordData<Vec<u8>, Name<Vec<u8>>>>;
-// type SomeRecord = ScannedRecord;
 type NameTypeTuple = (Name<Vec<u8>>, Rtype);
-
-// pub type ScannedDname = Chain<RelativeName<Bytes>, Name<Bytes>>;
-// pub type ScannedRecordData = ZoneRecordData<Bytes, ScannedDname>;
-// pub type ScannedRecord = Record<ScannedDname, ScannedRecordData>;
-// pub type ScannedString = Str<Bytes>;
 
 //------------ Update --------------------------------------------------------
 
@@ -101,8 +78,7 @@ pub struct Update {
     /// This specifies the prerequisite that at least one RR with a specified
     /// NAME and TYPE must exist.
     ///
-    /// Note that, if the domain name is relative, it will be relative to the
-    /// zone's apex.
+    /// If the domain name is relative, it will be relative to the zone's apex.
     #[arg(
         long = "rrset-exists",
         visible_alias = "rrset",
@@ -118,6 +94,8 @@ pub struct Update {
     /// This specifies the prerequisite that a set of RRs with a specified
     /// NAME and TYPE exists and has the same members with the same RDATAs as
     /// the RRset specified.
+    ///
+    /// If the domain name is relative, it will be relative to the zone's apex.
     #[arg(
         long = "rrset-exists-exact",
         visible_alias = "rrset-exact",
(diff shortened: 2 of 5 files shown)


src/commands/update.rs
```diff
@@ -1,60 +1,802 @@
...
+                };
+
+                let rcode = resp.header().rcode();
+                if rcode == Rcode::SERVFAIL || rcode == Rcode::NOTIMP {
+                    writeln!(env.stderr(), "Skipping _ @ {socket}: {rcode}");
+                    continue;
+                } else if rcode != Rcode::NOERROR {
+                    writeln!(env.stdout(), "UPDATE response was {rcode}");
+                }
+                return Ok(());
```
mozzieongit, line 714: Should we send the update to all name servers when they are provided by the users, or keep as is and stop after the first successful (like with the normal update procedure)?

src/commands/update.rs
```diff
@@ -1,60 +1,802 @@
...
+    /// Zone the domain name belongs to (to skip SOA query)
+    #[arg(short = 'z', long = "zone", value_name = "ZONE")]
     zone: Option<Name<Vec<u8>>>,
 
     /// TSIG credentials for the UPDATE packet
-    #[arg(short = 'y', long = "tsig", value_name = "name:key[:algo]")]
+    #[arg(short = 'y', long = "tsig", value_name = "NAME:KEY[:ALGO]")]
     tsig: Option<TSigInfo>,
+
+    /// RRset exists (value independent). (Optionally) provide this option
```
tertsdiepraam, line 74: The `(value independent)` part is a bit vague and the rest is a bit terse. Maybe something like this?
```
Require that an RRset exists before executing the update
```
Not great either, you can probably come up with something better.

src/commands/update.rs
```diff
@@ -1,60 +1,802 @@
...
+            } else if let Some(ttl) = arg.strip_suffix('m') {
+                ttl.parse::<u32>().map(|t| t / 60)
+            } else if let Some(ttl) = arg.strip_suffix('h') {
+                ttl.parse::<u32>().map(|t| t / 3600)
+            } else if let Some(ttl) = arg.strip_suffix('d') {
+                ttl.parse::<u32>().map(|t| t / 86400)
+            } else if let Some(ttl) = arg.strip_suffix('w') {
+                ttl.parse::<u32>().map(|t| t / 604800)
+            } else if let Some(ttl) = arg.strip_suffix('M') {
+                ttl.parse::<u32>().map(|t| t / 2629746) // 30.436875 days
```
tertsdiepraam, line 311: Is this a commonly known value? Is there a source saying that this is what people expect?

src/commands/update.rs
```diff
@@ -1,60 +1,802 @@
...
+                    update_section
+                        .push(Self::create_rr_addition(
+                            self.domain.clone(),
+                            self.class,
+                            self.ttl,
+                            rdata,
+                        ))
+                        .map_err(|e| -> Error {
+                            format!("Failed to add RR to UPDATE message: {e}").into()
+                        })?
```
tertsdiepraam, line 365: With this pattern it should be possible to do

```suggestion
                        .map_err(|e| {
                            format!("Failed to add RR to UPDATE message: {e}")
                        })?
```
But it's not super important.

src/commands/update.rs
```diff
@@ -1,60 +1,802 @@
...
+                }
+            }
+            UpdateAction::Clear => update_section
+                .push(Self::create_all_rrset_deletion(self.domain.clone()))
+                .map_err(|e| -> Error {
+                    format!("Failed to add RRset deletion RR to UPDATE message: {e}").into()
+                })?,
+        }
+
+        // Providing additional data is not yet implemented
```
tertsdiepraam, line 394:
```suggestion
        // TODO: Providing additional data is not yet implemented
```

src/commands/mod.rs
```diff
@@ -74,6 +74,13 @@ pub enum Command {
     #[command(name = "update")]
     Update(self::update::Update),
 
+    /// Send an UPDATE packet ldns compatibility variant
+    ///
+    /// This variant is not a dnst command and only used to provide
+    /// a separate implementation from the dnst update command.
+    #[command(skip)]
+    LdnsUpdate(self::update::LdnsUpdate),
```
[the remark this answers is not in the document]
mozzieongit: Yes. It is different because the Update struct now completely differs between `ldns-update` and `dnst update`.

src/commands/update.rs
```diff
@@ -1,60 +1,802 @@
...
 
-    /// Zone to update
-    #[arg(long = "zone")]
+    /// Class
+    #[arg(short = 'c', long = "class", default_value_t = Class::IN)]
+    class: Class,
+
+    /// TTL in seconds or with unit suffix (s, m, h, d, w, M, y).
+    #[arg(short = 't', long = "ttl", value_parser = Update::parse_ttl, default_value = "3600")]
+    ttl: Ttl,
```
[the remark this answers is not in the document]
mozzieongit: I left it on the base command to make it less confusing where options go. But I'll add a note to the docs/help.

src/commands/update.rs
```diff
@@ -1,60 +1,802 @@
...
+        //     }
+        // };
+
+        // Parse prerequisites before fetching nameservers, in case parsing fails
+        let prerequisites = self.parse_prerequisites(apex.clone().flatten_into())?;
+
+        // 1. (Cont.) If not provided, determine zone apex and fetch name servers
+        let nsnames = if self.nameservers.is_empty() {
+            Some(
+                // 3. Order nameserver list, listing primary first
```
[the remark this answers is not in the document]
mozzieongit: Step 2 is not implemented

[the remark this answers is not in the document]
tertsdiepraam: Sure!

src/commands/update.rs
mozzieongit: Maybe like so?
```
    /// Require that at least one RR with the given NAME and TYPE exists.
    /// (Optionally) provide this option multiple times, with format
    /// "<DOMAIN_NAME> <TYPE>" each, to build up a list of RRs.
    ///
    /// If the domain name is relative, it will be relative to the zone's apex.
...
    /// Require that an RRset exists and contains exactly the RRs with the given
    /// NAME, TYPE, and RDATA. (Optionally) provide this option multiple times,
    /// each with one RR in zonefile format, to build up one or more RRsets that
    /// is required to exist. CLASS and TTL can be omitted.
    ///
    /// If the domain name is relative, it will be relative to the zone's apex.
```

src/commands/update.rs
mozzieongit: It's the average month as a 12th of a year including leap time: `365.2425/12`
I was considering doing 30 or 31 days, or 4 weeks but couldn't decide, so went with average time instead.

mozzieongit: What would you expect one month to be?

mozzieongit: Also... uhm... 164f2607 ^^

mozzieongit: Given the hint about TTLs rarely being longer than a day, I removed the week, month, and year suffix.

tertsdiepraam (approved):
I'm a bit hesitant to give this an approve just because I'm not sure what it is supposed to do in each case, but the code looks great and well-documented! So on that basis I'll say LGTM!

src/commands/update.rs
```diff
@@ -1,60 +1,1228 @@
...
+    }
+
+    fn parse_prerequisite_name(
+        args: &Vec<String>,
+        origin: &Name<Vec<u8>>,
+    ) -> Result<Vec<Name<Vec<u8>>>, Error> {
+        let mut names = Vec::new();
+        for name in args {
+            let uncertain = UncertainName::<Vec<u8>>::from_str(name)
+                .map_err(|e| -> Error { format!("Invalid domain name '{name}': {e}").into() })?;
```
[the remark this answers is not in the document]
mozzieongit: I'll create a new PR for adding the tests and remove the `-> Error` things there. Then we have this merged in time for any actual release of dnst later.

PR #138, opened 2025-11-14
Introduce a 'WorkSpace' object to keep the current working state of k…

Philip-NLnetLabs:
…eyset.


bal-e (approved):
Looks good! Glad to see the simplifications to argument passing and the switch to `domain-kmip` :)

PR #139, opened 2025-11-19
Keyset locking

### LINKED ISSUE #150: keyset file locking
Keyset should use file locking to avoid concurrent access by multiple invocations of key on the same config file.

Philip-NLnetLabs:
Fixes #150.


src/commands/keyset/cmd.rs
```diff
@@ -3,19 +3,14 @@
...
-//   - option: update the config file atomically. This makes locking more
-//     complex.
 // - add a autoremove_delay option. Delete stale keys only some time after
 //   their withdrawn times.
 // - move dnskey_rrset, cds_rrset, ns_rrset into apex_extra (ds_rrset remains).
 //   add apex_remove with a list of Rtype that the signer should remove from
 //   the apex.
 // - add a -v option to keyset. Remove the -v option from status. Add
 //   verbose output for creation and deletion of keys.
+// - add a way for status and other read commands to by-pass the lock.
```
ximon18: Presumably that would only be safe if state updates are written atomically. Is that the case?

Cargo.toml
```diff
@@ -68,6 +68,8 @@ tracing = "0.1.41"
 tracing-subscriber = "0.3.19"
 url = "2.5.4"
 futures = "0.3.31"
+fs2 = "0.4.3"
```
ximon18, line 71: I'm wondering if this is a good choice of dependency. What were your reasons for selecting it?
I ask because:

- Krill uses https://lib.rs/crates/fd-lock.
- fs2 has a single maintainer.
- That maintainer hasn't responded to issues in the GH repo since 2017 as far as I can see.
- fs2 hasn't had a release in 8 years (though if it "just works" it may not need any).
- https://stackoverflow.com/questions/27469928/is-there-file-locking-in-rust#comment132066314_45040455 says:
  - (a) Rust has https://doc.rust-lang.org/stable/std/fs/struct.File.html#method.lock and,
  - (b) that there is more maintained fork of fs2 at https://github.com/al8n/fs4-rs.
- Conversely, I see that fd2 seems to be the most widely used file locking crate of some I locked at (fs2, fd-lock and file-lock).

src/commands/keyset/cmd.rs
```diff
@@ -5658,6 +5699,46 @@ fn show_automatic_roll_state(
     }
 }
 
+/// Open filename, get an exclusive lock and return the open file.
+///
+/// Assume changes are saved by creating a new file and renaming. After
+/// locking the file, the function has to check if the locked file is this
+/// the current fine under that name.
```
ximon18: This comment has some typos and grammatical issues.

src/commands/keyset/cmd.rs
```diff
@@ -5658,6 +5699,46 @@ fn show_automatic_roll_state(
...
+fn write_locked_file(filename: &PathBuf) -> Result<File, Error> {
+    // The config file is updated by writing to a new file and the renaming.
+    // Might have locked the old file. Check. Try a number of times and
+    // then give up. Lock contention is expected to be low.
+    for _try in 0..MAX_FILE_LOCK_TRIES {
```
ximon18, line 5689: Is it okay to retry lock acquisition with no delay between attempts?

src/commands/keyset/cmd.rs
```diff
@@ -5658,6 +5699,46 @@ fn show_automatic_roll_state(
     }
 }
 
+/// Open filename, get an exclusive lock and return the open file.
+///
+/// Assume changes are saved by creating a new file and renaming. After
+/// locking the file, the function has to check if the locked file is this
+/// the current fine under that name.
+fn write_locked_file(filename: &PathBuf) -> Result<File, Error> {
+    // The config file is updated by writing to a new file and the renaming.
```
ximon18:
```suggestion
    // The config file is updated by writing to a new file and then renaming.
```

src/commands/keyset/cmd.rs
```diff
@@ -5658,6 +5699,46 @@ fn show_automatic_roll_state(
...
+    // Might have locked the old file. Check. Try a number of times and
+    // then give up. Lock contention is expected to be low.
+    for _try in 0..MAX_FILE_LOCK_TRIES {
+        let file = File::open(filename)
+            .map_err(|e| format!("unable to open file {}: {e}", filename.display()))?;
+
+        file.lock_exclusive()
```
ximon18, line 5693: According to the documentation this call will block if the file is currently locked, is that desirable or would it be better to call `try_lock_exclusive()` instead?

src/commands/keyset/cmd.rs
```diff
@@ -5658,6 +5699,46 @@ fn show_automatic_roll_state(
...
+        let file_clone = file
+            .try_clone()
+            .map_err(|e| format!("unable to clone locked file {}: {e}", filename.display()))?;
+        let locked_file_handle = Handle::from_file(file_clone).map_err(|e| {
+            format!(
+                "Unable to get handle from locked file {}: {e}",
+                filename.display()
+            )
+        })?;
+        let current_file_handle = Handle::from_path(filename)
```
ximon18, line 5705: You open a file by path, lock it and get its handle, then get another handle for the same path and compare the two handles. Why? What is this handle comparison achieving/intended to do?

src/commands/keyset/cmd.rs
```diff
@@ -3760,6 +3742,65 @@ impl WorkSpace {
         let new_algs = HashSet::from([self.config.algorithm.to_generate_params().algorithm()]);
         curr_algs != new_algs
     }
+
+    /// Write config to a file.
+    ///
+    /// First write to a new filename and then rename to make sure that
+    /// changes are atomic.
+    fn write_config(&self, keyset_conf: &PathBuf) -> Result<(), Error> {
```
ximon18, line 3747: This and `fn write_state()` look almost identical, I would suggest factoring their core logic out into a single helper function, especially as they both contain the same comment and work around for lacking `-fn add_extension()` support.

src/commands/keyset/cmd.rs
```diff
@@ -3760,6 +3742,65 @@ impl WorkSpace {
...
+        // Rust 1.91.0 and above. Use strings instead.
+        // if !conf_file_new.add_extension("new") {
+        //	return Err(format!("unable to add extension 'new' to {}",
+        //		keyset_conf.display()).into());
+        // }
+        conf_file_new.as_mut_os_string().push(".new");
+
+        let mut file = File::create(&conf_file_new)
+            .map_err(|e| format!("unable to create file {}: {e}", conf_file_new.display()))?;
+        write!(file, "{json}")
```
ximon18: If some reason creation, writing or renaming fail, the created `.new` file will be left behind, right? Is that a problem? Subsequent attempts would fail I assume as the created file already exists. Should you attempt to remove the file in such scenarios?

src/commands/keyset/cmd.rs
```diff
@@ -5658,6 +5699,46 @@ fn show_automatic_roll_state(
     }
 }
 
+/// Open filename, get an exclusive lock and return the open file.
+///
+/// Assume changes are saved by creating a new file and renaming. After
+/// locking the file, the function has to check if the locked file is this
+/// the current fine under that name.
+fn write_locked_file(filename: &PathBuf) -> Result<File, Error> {
```
ximon18: I find the name of this function confusing, as "write" reads like a verb yet this function does neither opens the file for writing nor writes to it.

ximon18 (approved):

src/commands/keyset/cmd.rs
Philip-NLnetLabs: That is the case because the state is written to a new file and then renamed. There is a small risk that config and state are not in sync. This list is a TODO list. I'll move them to issues.

src/commands/keyset/cmd.rs
Philip-NLnetLabs: This will lead to the thundering herd problem. But concurrent access to keyset for a single domain is expected to be very low. We need to do something if we ever fail to get a lock within MAX_FILE_LOCK_TRIES.

src/commands/keyset/cmd.rs
Philip-NLnetLabs: For this PR the goal is to block. Maybe the UI needs to be improved in the future.

src/commands/keyset/cmd.rs
Philip-NLnetLabs: Updated to:
```
/// Assume changes are saved by creating a new file and renaming. After
/// locking the file, the function has to check if the locked file is the
/// same as the current file under that name.
```

src/commands/keyset/cmd.rs
Philip-NLnetLabs: This is explained by the comment at the top of the function. Maybe it needs more text. The idea is that because the config is saved by creating a new file and then renaming that file, it is possible to endup with a lock on the old config file that no longer exists. So check that we have locked to current config file.

ximon18: You only call `write_locked_file()` once, so any rename that happens within your code when the config file is saved won't be detected later by this logic, at least currently. So is this for the some future potential second call to this function after saving/renaming of the config file?

Also, when this test fails you call `continue` so you go round the loop again, dropping the file you just opened and the lock you just acquired, and try re-opening the file by the same filename as you just tried, even though that succeeded. I don't understand what this achieves.

If this function were called while a second concurrent invocation of `dnst keyset` was busy saving the config file, one would be able to open the file then block until the lock can be acquired, is this the issue, that you would then be possibly be opening the old FD? (though what happens to that FD during the rename/once the rename completes, is it still valid?)

src/commands/keyset/cmd.rs
Philip-NLnetLabs: Done.

src/commands/keyset/cmd.rs
Philip-NLnetLabs: I don't expect create to fail if a file with the same name already exists.

src/commands/keyset/cmd.rs
Philip-NLnetLabs: Changed to file_with_write_lock

Cargo.toml
Philip-NLnetLabs: I'll leave it at fs2 for now. If fs2 gets less popular or picks up security issues then we can switch. The interface of fd-lock looks scary.

src/commands/keyset/cmd.rs
Philip-NLnetLabs: After a rename, any processes that run keyset and that were trying to lock the config file now try to lock the wrong file. So they need to start again. The logic of keyset is that it gets a lock on the config file, and then saves the state file if it changed and saves the config file if it changed. There will not be a second call to lock the config file in a single run of keyset.

Issue #53, opened 2025-01-10
Support autocompletion of algorithm names for `dnst keygen` argument `-a`.


ximon18:

mozzieongit:
This seems to be a bit tricky. 

Using clap_complete, we could use PossibleValues to provide a list of strings possible for completion. Unfortunately, when doing so, clap only accepts those strings as possible values. There seems to be no option to have these possible values just as a basis for completion, but allow more values as inputs. So, if we were to provide a list of algorithms to complete (in keygen):
```rust
// ...
value_parser = clap::builder::PossibleValuesParser::new([
    "RSASHA256", "ECDSAP256SHA256", "ECDSAP384SHA384", "ED25519", "ED448", "8", "13", "14", "15", "16"
]).map(|s| Keygen::parse_algorithm(&s).unwrap()),
// ...
algorithm: GenerateParams,
```
then clap wouldn't allow any other inputs like `RSASHA256:<bits>`. So we could change that to make the bits argument a separate option, or:

There is work done on dynamic completions with clap, but it's not stabilized yet (and doesn't seem to work on my first tests):
- https://github.com/clap-rs/clap/issues/3166
- https://docs.rs/clap_complete/latest/clap_complete/engine/struct.ArgValueCompleter.html

Another option would be to generate the completion by clap and add the algorithm completion by hand/script to the completion script.

Another point, by shell convention the bits parameter for RSA would be separated by a `=` so `RSASHA256=2048` instead of `RSASHA256:2048`, which would allow for specifying in the shell completion code that RSA takes an optional number argument.

---

General completion code generation exists in branch [shell-completion](https://github.com/NLnetLabs/dnst/tree/shell-completion).

---

Scratchpad for code snippets:

```
# e.g. in ZSH
$ dnst keygen --algorithm=<TAB>
ECDSAP256SHA256  13  -- ECDSAP256SHA256
ECDSAP384SHA384  14  -- ECDSAP384SHA384
ED25519          15  -- ED25519
ED448            16  -- ED448
RSASHA256        8   -- RSASHA256:<bits> with optional key size
$ dnst keygen --algorithm=RSA<TAB>
# results in
$ dnst keygen --algorithm=RSASHA256=

# snippet from ZSH completion code
(keygen)
	__algorithms() {
		_values algorithm \
			'RSASHA256[RSASHA256:<bits> with optional key size]::bits:' \
			'ECDSAP256SHA256[ECDSAP256SHA256]' \
[...]

PR #144, opened 2025-11-28
Change keygen RSA bits argument into its own option

mozzieongit:
Originally, I found a discrepancy between the dnst-keygen man page (listing the -b option) and the code. 
Instead of just fixing the docs I remembered hacking around for #53 and struggling because of the combination of the bits parameter with the algorithm selection.
By splitting the bits parameter off again, generating auto completion will be possible using clap without any special hacks or manual insertion of manually written shell completion code.

### DIFF
--- a/src/commands/keygen.rs
+++ b/src/commands/keygen.rs
@@ -23,6 +23,8 @@ use super::{parse_os, parse_os_with, Command, LdnsCommand};
 #[cfg(not(any(feature = "openssl", feature = "ring")))]
 compile_error!("Either the 'openssl' or the 'ring' feature (or both) must be enabled");
 
+const DEFAULT_RSA_KEYLEN: u32 = 2048;
+
 #[derive(Clone, Debug, PartialEq, Eq, clap::Args)]
 pub struct Keygen {
     /// The signature algorithm to generate for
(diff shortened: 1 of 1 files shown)


src/commands/keygen.rs
```diff
@@ -228,17 +237,7 @@ impl LdnsCommand for Keygen {
...
-            Some(SecurityAlgorithm::ECDSAP256SHA256) => GenerateParams::EcdsaP256Sha256,
-            Some(SecurityAlgorithm::ECDSAP384SHA384) => GenerateParams::EcdsaP384Sha384,
-            Some(SecurityAlgorithm::ED25519) => GenerateParams::Ed25519,
-            Some(SecurityAlgorithm::ED448) => GenerateParams::Ed448,
-            Some(_) => unreachable!(),
-            None => {
-                return Err("Missing algorithm (-a) option".into());
-            }
-        };
+        let algorithm = algorithm.ok_or("Missing algorithm (-a) option")?;
```
bal-e: Shouldn't we mark the option as `required` in Clap instead?

bal-e (commented):
I agree that this CLI change would simplify the interface and better support auto-complete.  But I have a few concerns:

- Can we make sure to return an error if `-b` is specified without the RSA algorithm?
- How will this affect Cascade?  Do we need to coordinate PRs, and update the version compatibility check ?

bal-e:
Yeah, my concern is that users will add `-b` for a non-RSA algorithm and miss that it's irrelevant.  Good to know that Cascade is unaffected.

PR #146, opened 2025-12-16
Keyset faketime

Philip-NLnetLabs:
Add fake-time to supported testing. This works fine for generating signatures with specific inception and expiration times. At the moment keyset in domain has no support for faketime. So key rolls cannot be tested with virtual time.


ximon18 (commented):
Shouldn't `sig_renew()` also use fake-time?

doc/manual/build/man/dnst-key2ds.1
```diff
@@ -27,7 +27,7 @@ level margin: \\n[rst2man-indent\\n[rst2man-indent-level]]
 .\" new: \\n[rst2man-indent\\n[rst2man-indent-level]]
 .in \\n[rst2man-indent\\n[rst2man-indent-level]]u
 ..
-.TH "DNST-KEY2DS" "1" "Feb 25, 2026" "0.1.1-dev" "dnst"
+.TH "DNST-KEY2DS" "1" "Feb 16, 2026" "0.1.1-dev" "dnst"
```
ximon18: Maybe regenerate the files to prevent this metadata going backward in time?

src/commands/keyset/cmd.rs
```diff
@@ -1678,6 +1710,11 @@ struct KeySetConfig {
 
     /// Command to run when the DS records at the parent need updating.
     update_ds_command: Vec<String>,
+
+    /// Fake time to use when signing.
+    ///
+    /// This is need for integration tests.
```
ximon18:
```suggestion
    /// This is needed for integration tests.
```

src/commands/keyset/cmd.rs
ximon18: Should this also use fake time?

src/commands/keyset/cmd.rs
```diff
@@ -2882,9 +2924,9 @@ impl WorkSpace {
                 dnskeys.push(public_key);
             }
         }
-        let now = Timestamp::now().into_int();
-        let inception = (now - self.config.dnskey_inception_offset.as_secs() as u32).into();
-        let expiration = (now + self.config.dnskey_signature_lifetime.as_secs() as u32).into();
+        let now_u32 = Into::<Duration>::into(now).as_secs() as u32;
```
ximon18, line 2909: Since there is no other use of `now` in this function below the use of `now_u32` here, why create a second `now_u32` variable instead of replacing the existing `now` variable?

src/commands/keyset/cmd.rs
```diff
@@ -3025,9 +3068,9 @@ impl WorkSpace {
             // Need to sign
         }
 
-        let now = Timestamp::now().into_int();
-        let inception = (now - self.config.cds_inception_offset.as_secs() as u32).into();
-        let expiration = (now + self.config.cds_signature_lifetime.as_secs() as u32).into();
+        let now_u32 = Into::<Duration>::into(now).as_secs() as u32;
```
ximon18, line 3053: Since there is no other use of `now` in this function below the use of `now_u32` here, why create a second `now_u32` variable instead of replacing the existing `now` variable?

src/commands/keyset/cmd.rs
```diff
@@ -2865,6 +2906,7 @@ impl WorkSpace {
     /// Collect all keys where present() returns true and sign the DNSKEY RRset
     /// with all KSK and CSK (KSK state) where signer() returns true.
     fn update_dnskey_rrset(&mut self, env: &impl Env, verbose: bool) -> Result<(), Error> {
+        let now = self.faketime_or_now();
```
ximon18, line 2891: In normal use (without fake time) this timestamp could differ to the "now" used elsewhere, but when using fake time those two timestamps will be the same I assume. Why is it okay/required for them to differ in normal use but not when using fake time?

src/commands/keyset/cmd.rs
```diff
@@ -4226,129 +4175,432 @@ async fn auto_wait_actions(
...
+            // need to be no-ops or report a TTL. Wait actions are not
+            // compatible with this.
+            Action::WaitDnskeyPropagated
+            | Action::WaitDsPropagated
+            | Action::WaitRrsigPropagated => unreachable!(),
+        }
+    }
+    AutoReportActionsResult::Report(max_ttl)
+}
+/// Create CDS and CDNSKEY RRsets.
```
ximon18, line 3995: Missing line break?

src/commands/keyset/cmd.rs
```diff
@@ -3760,363 +3811,257 @@ impl WorkSpace {
...
                 drop(report_state_locked);
                 *state_changed = true;
 
                 match result {
-                    AutoReportActionsResult::Wait(next) => return AutoActionsResult::Wait(next),
-                    AutoReportActionsResult::Report(_) => (),
+                    AutoReportActionsResult::Wait(_) => return result,
+                    AutoReportActionsResult::Report(ttl) => {
+                        max_ttl = max(max_ttl, ttl);
+                    }
```
ximon18: What does this change have to do with introducing fake time?

src/commands/keyset/cmd.rs
```diff
@@ -3760,363 +3811,257 @@ impl WorkSpace {
...
                 // Clippy problem
                 {
                     let report_state_locked = report_state.lock().expect("lock() should not fail");
                     if let Some(ds_status) = &report_state_locked.ds {
                         match ds_status {
                             AutoReportActionsResult::Wait(next) => {
-                                if *next > UnixTime::now() {
-                                    return AutoActionsResult::Wait(next.clone());
+                                if *next > now {
+                                    return ds_status.clone();
```
ximon18: What does this change have to do with introducing fake time?

src/commands/keyset/cmd.rs
```diff
@@ -3760,363 +3811,257 @@ impl WorkSpace {
...
                                 }
                             }
-                            AutoReportActionsResult::Report(_) => continue,
+                            AutoReportActionsResult::Report(ttl) => {
+                                max_ttl = max(max_ttl, *ttl);
```
ximon18: What does this change have to do with introducing fake time?

src/commands/keyset/cmd.rs
```diff
@@ -3760,363 +3811,257 @@ impl WorkSpace {
...
                 report_state_locked.ds = Some(result.clone());
                 drop(report_state_locked);
                 *state_changed = true;
 
                 match result {
-                    AutoReportActionsResult::Wait(next) => return AutoActionsResult::Wait(next),
-                    AutoReportActionsResult::Report(_) => (),
+                    AutoReportActionsResult::Wait(_) => return result,
+                    AutoReportActionsResult::Report(ttl) => {
+                        max_ttl = max(max_ttl, ttl);
```
ximon18: What does this change have to do with introducing fake time?

src/commands/keyset/cmd.rs
```diff
@@ -4130,34 +4075,41 @@ async fn auto_wait_actions(
...
                         AutoReportRrsigResult::Wait(next) => {
-                            if next > UnixTime::now() {
-                                return AutoActionsResult::Wait(next.clone());
+                            if next > now {
+                                return AutoReportActionsResult::Wait(next.clone());
                             }
                         }
-                        AutoReportRrsigResult::Report(_) => continue,
+                        AutoReportRrsigResult::Report(ttl) => {
+                            max_ttl = max(max_ttl, ttl);
```
ximon18: What does this change have to do with introducing fake time?

src/commands/keyset/cmd.rs
```diff
@@ -4130,34 +4075,41 @@ async fn auto_wait_actions(
...
                         AutoReportRrsigResult::WaitSoa {
                             next,
                             serial,
                             ttl,
                             report_ttl,
                         } => {
-                            if next > UnixTime::now() {
-                                return AutoActionsResult::Wait(next.clone());
+                            if next > now {
+                                return AutoReportActionsResult::Wait(next.clone());
```
ximon18: What does this change have to do with introducing fake time?

ximon18 (commented):
Looks mostly fine, am confused by some changes regarding TTLs that appear unrelated to fake time, is that some weird artifact of the GH diff view?

doc/manual/build/man/dnst-key2ds.1
Philip-NLnetLabs: I made them identical to the ones currently in main.

src/commands/keyset/cmd.rs
Philip-NLnetLabs: Yes, done.

src/commands/keyset/cmd.rs
Philip-NLnetLabs: To avoid confusion. now_u32 is a truncated version of now.

src/commands/keyset/cmd.rs
Philip-NLnetLabs: It is fine if they are different. The question is whether it would improve code if now was stored and the code would use self.now. Hard to estimate. I'll leave it like this for now.

Philip-NLnetLabs:
I think some of the weird changes are GH getting confused. One function got moved. I moved it back to its original location and the diffs are gone.

ximon18 (approved):

PR #149, opened 2026-01-23
Keyset extra KSK and ZSK rolls

### LINKED ISSUE #154: keyset additonal keyrolls
make additional key rolls available.

Philip-NLnetLabs:
Fixes #154.


src/commands/keyset/cmd.rs
```diff
@@ -1609,6 +1641,78 @@ struct AutoConfig {
...
+/// Type of KSK roll to perform.
+#[derive(Clone, Debug, Default, Deserialize, Serialize)]
+enum KskRollType {
+    #[default]
+    /// KSK roll that first adds the new KSK to the DNSKEY RRset and add an
+    /// additional signature over the DNSKEY RRset from that key before
+    /// replace the DS for the old key with one for the new key.
+    DoubleSignatureKskRoll,
+
+    /// KSK roll that first publishes an additional DS record for the new
```
ximon18, line 1756: For the new what?

src/commands/keyset/cmd.rs
```diff
@@ -1609,6 +1641,78 @@ struct AutoConfig {
...
+            KskRollType::DoubleDsKskRoll => write!(fmt, "double-ds-ksk-roll"),
+        }
+    }
+}
+
+/// Type of ZSK key roll to use.
+#[derive(Clone, Debug, Default, Deserialize, Serialize)]
+enum ZskRollType {
+    #[default]
+    /// Type of ZSK roll where the new ZSK is first added to the DNSKEY
```
ximon18: Inconsistent: In `KskRollType` above you don't have the leading "Type of" text.

ximon18 (commented):
As this PR doesn't seem to contain the logic to act on the new key roll types, but e.g. I see some existing logic relating to `ZskDoubleSignatureRoll`, is it the case that the logic for the additional key roll types already existed but was not selectable by the operator?

Philip-NLnetLabs:
Yes, all of the logic is in domain. The only code needed here is the config settings to select which roll variants to use.

ximon18 (approved):

PR #164, opened 2026-03-04
Remove ZONEMD records from the input zone.

### LINKED ISSUE #159: signzone should not copy ZONEMD records
Currently, signzone copies ZONEMD records from the input zone to the signed output zone when not adding a ZONEMD record. These copied ZONEMD records are not valid because they lack the correct digest. The correct solution is to remove them.

Philip-NLnetLabs:
Fixes #159.

### DIFF
--- a/src/commands/signzone.rs
+++ b/src/commands/signzone.rs
@@ -1464,7 +1464,15 @@ impl SignZone {
                     // release a fixed version of ldns-signzone that strips
                     // NSEC(3)s.
                     //
+                    // Remove ZONEMD records at apex as well. We don't always
+                    // know the origin at this point. Just strip all ZONEMD
+                    // records as they are currently only defined for
+                    // use at the apex.
+                    //
                     // TODO: Support partial and re-signing.
+                    if matches!(record.rtype(), Rtype::ZONEMD) {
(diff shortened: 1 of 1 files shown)


src/commands/signzone.rs
```diff
@@ -1464,7 +1464,24 @@ impl SignZone {
                     // release a fixed version of ldns-signzone that strips
                     // NSEC(3)s.
                     //
+                    // Remove ZONEMD records at apex as well. We don't always
+                    // know the origin at this point. Just strip all ZONEMD
+                    // records if we don't, strip ZONEMD records at apex
+                    // if we do know the origin.
+                    //
                     // TODO: Support partial and re-signing.
```
mozzieongit: Minor nit: Move the new comment under the TODO

```suggestion
                    // TODO: Support partial and re-signing.
                    //
                    // Remove ZONEMD records at apex as well. We don't always
                    // know the origin at this point. Just strip all ZONEMD
                    // records if we don't, strip ZONEMD records at apex
                    // if we do know the origin.
```

mozzieongit (approved):
Just a minor nit about the comment location. Otherwise LGTM

PR #167, opened 2026-04-02
Introduces apex_remove and apex_extra.

### LINKED ISSUE #152: keyset move all apex records into a single state variable
- move dnskey_rrset, cds_rrset, ns_rrset into apex_extra (ds_rrset remains).
- add apex_remove with a list of Rtypes that the signer should remove from the apex.

Philip-NLnetLabs:
Apex_remove and apex_extra will replace dnskey_rrset, cds_rrset, and ns_rrset. Fixes #152

### DIFF
--- a/src/commands/keyset/cmd.rs
+++ b/src/commands/keyset/cmd.rs
@@ -709,7 +709,7 @@ impl Keyset {
                 cds_rrset: Vec::new(),
                 ns_rrset: Vec::new(),
                 apex_remove: (*APEX_REMOVE).into(),
-                apex_records: Vec::new(),
+                apex_extra: Vec::new(),
                 cron_next: None,
                 internal: HashMap::new(),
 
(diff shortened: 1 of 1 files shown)


ximon18:
You don't say here or on #152 why this change is being made.

src/commands/keyset/cmd.rs
```diff
@@ -4073,7 +4093,20 @@ impl WorkSpace {
     }
 
     /// Write state to a file.
-    fn write_state(&self) -> Result<(), Error> {
+    fn write_state(&mut self) -> Result<(), Error> {
+        // Always set apex_remove.
+        self.state.apex_remove = (*APEX_REMOVE).into();
+
+        // Update apex_extra from the old fields.
```
ximon18, line 4100: This looks like it doesn't update `apex_extra` but overwrites it. Is that correct?

Philip-NLnetLabs: Yes. In the future the old fields need to be removed and then apex_extra gets updated in the right place. At the moment just overwriting is best.

Philip-NLnetLabs:
apex_remove is introduced to avoid hardcoding in signers what records to remove from the unsigned zone. In particular, a key manager with multi-signer support may take control over NS. apex_extra is introduced to avoid having multiple fields that all need to be added to the apex. With the question what has to happen when the key manager need to deal with new types. For example CSYNC.

ximon18 (approved):

PR #171, opened 2026-05-07
keygen, keyset: use secure file permissions for key material

### LINKED ISSUE #238: Set file permissions for key material to read/write by cascade user only when creating the files
With 0.1.0-alpha2:

```
% ls -la /var/db/cascade/keys/
total 64
drwxr-xr-x 2 cascade users   12288 Oct 17 14:55 .
drwxr-xr-x 3 cascade root     4096 Sep 19 16:34 ..
-rw-rw-rw- 1 cascade cascade    82 Oct 17 14:51 Kinternautique.fr.+015+27737.key
-rw-rw-rw- 1 cascade cascade   106 Oct 17 14:51 Kinternautique.fr.+015+27737.private
```

maertsen:
(private: u=rw,go= and public: u=rw,og=r)

Reported-by: Stéphane Bortzmeyer
Fixes: https://github.com/NLnetLabs/cascade/issues/238

### DIFF
--- a/src/commands/keygen.rs
+++ b/src/commands/keygen.rs
@@ -683,7 +683,6 @@ mod test {
                 "Unsafe permissions on private key"
             );
         }
-
     }
 
     #[test]
--- a/src/commands/keygen.rs
+++ b/src/commands/keygen.rs
@@ -350,11 +350,11 @@ impl Keygen {
         let public_key_path = format!("{base}.key");
         let digest_file_path = self.make_ksk.then(|| format!("{base}.ds"));
 
-        let mut secret_key_file = util::create_new_file(&env, &secret_key_path)?;
-        let mut public_key_file = util::create_new_file(&env, &public_key_path)?;
+        let mut secret_key_file = util::create_new_file(&env, &secret_key_path, 0o600)?;
+        let mut public_key_file = util::create_new_file(&env, &public_key_path, 0o644)?;
         let mut digest_file = digest_file_path
             .as_ref()
-            .map(|digest_file_path| util::create_new_file(&env, digest_file_path))
+            .map(|digest_file_path| util::create_new_file(&env, digest_file_path, 0o644))
             .transpose()?;
 
         Self::symlink(&secret_key_path, ".private", self.symlink, &env)?;
@@ -431,6 +431,7 @@ impl Keygen {
 mod test {
     use domain::base::iana::SecurityAlgorithm;
     use regex::Regex;
+    use std::os::unix::fs::PermissionsExt;
 
     use crate::commands::Command;
     use crate::env::fake::FakeCmd;
@@ -667,6 +668,17 @@ mod test {
         let secret_key =
             std::fs::read_to_string(dir.path().join(format!("{name}.private"))).unwrap();
         assert!(secret_key_regex.is_match(&secret_key));
+
+        let permissions = std::fs::File::open(dir.path().join(format!("{name}.private")))
+            .unwrap()
+            .metadata()
+            .unwrap()
+            .permissions();
+        assert_eq!(
+            permissions.mode() & 0o777,
+            0o600,
+            "Unsafe permissions on private key"
+        );
     }
 
     #[test]
--- a/src/commands/keygen.rs
+++ b/src/commands/keygen.rs
@@ -431,6 +431,7 @@ impl Keygen {
 mod test {
     use domain::base::iana::SecurityAlgorithm;
     use regex::Regex;
+    #[cfg(unix)]
     use std::os::unix::fs::PermissionsExt;
 
     use crate::commands::Command;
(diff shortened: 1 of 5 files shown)


maertsen:
So, `mode()` is defined by the `OpenOptionsExt` trait, which is not available on Windows. If I were to make that optional, we would (presumably) retain the unsafe behaviour there, which is not great.
But, if this gets merged as is, to fix the behaviour for Cascade, it breaks builds of dnst for Windows going forward.

Not sure how to proceed. From some light searching, I understand the Windows side is not covered in the stdlib.

src/commands/keyset/cmd.rs
```diff
@@ -60,6 +60,7 @@ use std::fmt::{Debug, Display, Formatter};
 use std::fs::{create_dir_all, remove_file, rename, File};
 use std::io::{self, Write};
 use std::net::{IpAddr, SocketAddr};
+use std::os::unix::fs::OpenOptionsExt;
```
mozzieongit, line 64:
```suggestion
#[cfg(unix)]
use std::os::unix::fs::OpenOptionsExt;
```

src/commands/keyset/cmd.rs
```diff
@@ -4272,7 +4273,12 @@ impl WorkSpace {
         //		ws.config.state_file.display()).into());
         // }
         filename_new.as_mut_os_string().push(".new");
-        let mut file = File::create(&filename_new)
+        let mut file = File::options()
+            .write(true)
+            .create(true)
+            .truncate(true)
+            .mode(0o644)
```
mozzieongit:
```suggestion
        let mut file_opts = File::options();
        file_opts.write(true).create(true).truncate(true);
        #[cfg(unix)]
        file_opts.mode(0o644);
        let mut file = file_opts
```

src/commands/keygen.rs
```diff
@@ -431,6 +431,7 @@ impl Keygen {
 mod test {
     use domain::base::iana::SecurityAlgorithm;
     use regex::Regex;
+    use std::os::unix::fs::PermissionsExt;
```
mozzieongit, line 435:
```suggestion
    #[cfg(unix)]
    use std::os::unix::fs::PermissionsExt;
```

src/commands/keygen.rs
```diff
@@ -667,6 +668,17 @@ mod test {
...
+        let permissions = std::fs::File::open(dir.path().join(format!("{name}.private")))
+            .unwrap()
+            .metadata()
+            .unwrap()
+            .permissions();
+        assert_eq!(
+            permissions.mode() & 0o777,
+            0o600,
+            "Unsafe permissions on private key"
+        );
```
mozzieongit:
```suggestion
        #[cfg(unix)]
        {
            let permissions = std::fs::File::open(dir.path().join(format!("{name}.private")))
                .unwrap()
                .metadata()
                .unwrap()
                .permissions();
            assert_eq!(
                permissions.mode() & 0o777,
                0o600,
                "Unsafe permissions on private key"
            );
        }

```

src/util.rs
```diff
@@ -1,16 +1,22 @@
 //! A utility module for common operations.
 
 use std::fs::File;
+use std::os::unix::fs::OpenOptionsExt;
```
mozzieongit, line 5:
```suggestion
#[cfg(unix)]
use std::os::unix::fs::OpenOptionsExt;
```

src/util.rs
```diff
@@ -1,16 +1,22 @@
...
-pub fn create_new_file(env: &impl Env, path: impl AsRef<Path>) -> Result<File> {
+pub fn create_new_file(env: &impl Env, path: impl AsRef<Path>, mode: u32) -> Result<File> {
     let path = path.as_ref();
     let abs_path = env.in_cwd(&path);
-    File::create_new(abs_path)
+    File::options()
+        .mode(mode)
+        .read(true)
+        .write(true)
+        .create_new(true)
```
mozzieongit:
```suggestion
    let mut file_opts = File::options();
    file_opts.read(true).write(true).create_new(true);
    #[cfg(unix)]
    file_opts.mode(mode);
    file_opts
```

mozzieongit:
> So, `mode()` is defined by the `OpenOptionsExt` trait, which is not available on Windows. 

> If I were to make that optional, we would (presumably) retain the unsafe behaviour there, which is not great. But, if this gets merged as is, to fix the behaviour for Cascade, it breaks builds of dnst for Windows going forward.

I would make setting the mode optional (aka use the `cfg(unix)`).
This way we don't break the build for windows, but get the fix for cascade.

Keeping the old behaviour in windows is not really an issue, I think, as a windows user will probably not use keyset (because that is mainly intended for cascade, which is linux only). A windows user might use keygen, but then it's an active step by the user and we could reasonably expect them to protect their key files themselves.

I added the necessary lines as review comments for you, because figuring out the cfg stuff is a bit annoying in the beginning.

src/util.rs
```diff
@@ -1,16 +1,22 @@
...
 use std::fs::File;
+use std::os::unix::fs::OpenOptionsExt;
 use std::path::Path;
 
 use crate::env::Env;
 use crate::error::Result;
 
 /// Create and open a file.
-pub fn create_new_file(env: &impl Env, path: impl AsRef<Path>) -> Result<File> {
+pub fn create_new_file(env: &impl Env, path: impl AsRef<Path>, mode: u32) -> Result<File> {
```
mozzieongit:
```suggestion
pub fn create_new_file(
    env: &impl Env,
    path: impl AsRef<Path>,
    #[cfg_attr(not(unix), allow(unused_variables))] mode: u32,
) -> Result<File> {
```

Oh, and because we will only use the mode parameter on unix, we need to allow it to be unused on not-unix

mozzieongit (approved):

PR #184, opened 2026-07-31
Document the default values for `keyset set` subcommands.

ximon18:
This started out as an attempt to document the default KSK and ZSK roll types and expanded to cover documenting the defaults for all `keyset set` settable values.

With regard to key roll types, one open question I have is whether the references I have added to RFC 7583 key roll type names are correct and wanted in the docs.

### DIFF
--- a/doc/manual/source/man/dnst-keyset.rst
+++ b/doc/manual/source/man/dnst-keyset.rst
@@ -717,12 +717,12 @@ The keyset subcommand provides the following commands:
   * ksk-roll-type <KSK-ROLL-TYPE>
 
     The type of KSK roll to use. Possible values are double-signature-ksk-roll
-    and double-ds-ksk-roll.
+    and double-ds-ksk-roll. Defaults to double-signature-ksk-roll.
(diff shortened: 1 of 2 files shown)


doc/manual/source/man/dnst-keyset.rst
```diff
@@ -732,21 +745,26 @@ The keyset subcommand provides the following commands:
 
     For example, ``auto-csk true false true false`` means that
     CSK rolls will start automatically, that the propagation1-complete,
-    propagation2-complete, and roll-done need to be executed manually.
+    propagation`2-complete, and roll-done need to be executed manually.
```
Philip-NLnetLabs: Why is there an extra quote character?

ximon18: User error, sorry, will fix.

ximon18: Removed.

Philip-NLnetLabs (approved):

